PR #5, opened 2019-07-05
Returns the number of surfaces and volumes in a file

ep09772:
Added Entityset into the types list and separated out the volumes and surfaces based on their tags. Created a type key so that below the counts of each entitytype each number has the corresponding label (i.e. Type 0: Vertex).

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -51,3 +51,4 @@ def get_entity_ranges(my_core, meshset, entity_types):
 
     return entity_ranges
 
+get_dagmc_tags(core.Core())
--- a/generate_stats.py
+++ b/generate_stats.py
@@ -17,10 +17,9 @@ def report_stats(entity_ranges):
     ------
     entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
     """
-
     for entity_type, eh_range in entity_ranges.items():
         print("There are {} entities of type {}.".format(eh_range.size(),entity_type))
-
+    print('Type 0: Vertices \nType 2: Triangles \nType 11: EntitySets')
 def main():
 
     # starting with a single input file - will need to convert this to a user option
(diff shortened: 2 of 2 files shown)


dagmc_stats.py
```diff
@@ -51,3 +51,4 @@ def get_entity_ranges(my_core, meshset, entity_types):
 
     return entity_ranges
 
+get_dagmc_tags(core.Core())
```
gonuke: Not sure why this line is here in the `dagmc_stats` module.  This line will be run every time, and `core.Core()` will be an empty instance of MOAB with no tags

generate_stats.py
```diff
@@ -17,10 +17,9 @@ def report_stats(entity_ranges):
     ------
     entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
     """
-
     for entity_type, eh_range in entity_ranges.items():
         print("There are {} entities of type {}.".format(eh_range.size(),entity_type))
-
+    print('Type 0: Vertices \nType 2: Triangles \nType 11: EntitySets')
```
gonuke: This is a useful legend, but might be helpful to  implement a way to provide this information instead of the entity number in the previous print statement.

generate_stats.py
```diff
@@ -32,11 +31,10 @@ def main():
 
     # get tags
     dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
-    
+    print(dagmc_tags)
```
gonuke: Is this for debugging or part of the normal output that you expect?

dagmc_stats.py
ep09772: This isn't supposed to be here, it got deleted somewhere later.

generate_stats.py
ep09772: Okay, I'll implement that in

generate_stats.py
ep09772: Debugging, it is also gone in more recent versions

PR #8, opened 2019-07-05
Returns EntitySet Ranges

ep09772:
Added comments to wrap up the function

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -1,17 +1,16 @@
 # set the path to find the current installation of pyMOAB
 import sys
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
+import numpy as np
 from pymoab import core, types
 from pymoab.rng import Range
-
 def get_dagmc_tags(my_core):
     """
     Get a dictionary with the important tags for DAGMC geometries
-
+    
     inputs
     ------
     my_core : a MOAB Core instance
-
     outputs
     -------
     dagmc_tags : a dictionary of relevant tags
@@ -19,7 +18,7 @@ def get_dagmc_tags(my_core):
 
     dagmc_tags = {}
     dagmc_tags['geom_dim'] = my_core.tag_get_handle('GEOM_DIMENSION', size = 1, tag_type = types.MB_TYPE_INTEGER, #creates tag for the geo-
-                                                 storage_type = types.MB_TAG_SPARSE, create_if_missing = True)# metric dimension
+                                                 storage_type = types.MB_TAG_SPARSE, create_if_missing = True)# geometric dimension
     
     dagmc_tags['category'] = my_core.tag_get_handle('CATEGORY', size = 32, tag_type = types.MB_TYPE_OPAQUE, #creates tag for the word of 
                                                  storage_type = types.MB_TAG_SPARSE, create_if_missing = True) #the category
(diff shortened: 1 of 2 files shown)


dagmc_stats.py
```diff
@@ -46,8 +46,11 @@ def get_entity_ranges(my_core, meshset, entity_types):
     """
 
     entity_ranges = {}
-    for entity_type in entity_types:
+    for entity_type in entity_types:#goes through each entitytype given
         entity_ranges[entity_type] = my_core.get_entities_by_type(meshset, entity_type) 
-
+        if entity_type == 11: #if the type is mbentitytypes
```
gonuke: I’d prefer to see this capability implemented in a different function:
- there is a semantic difference between the raw MOAB entities/entitySets and these tagged/categorized sets
- there is a similar semantic difference in the keys to the `entity_ranges` dictionary

generate_stats.py
```diff
@@ -32,11 +31,10 @@ def main():
 
     # get tags
     dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
-    
+    print(dagmc_tags)
```
gonuke: I think this is just from debugging, please delete.

There may be value in adding a method to write a summary of the tags.

gonuke: But writing out info about tags should probably be a different PR

gonuke:
I think this is a good narrow-scope PR to focus on getting done to practice the process of updating and reviewing PRs.  A PR is like a writing assignment that ends up with a few rounds of editing and review, whether by you or others.

You may want to review [PEP8](https://www.python.org/dev/peps/pep-0008/) that we use for our style guide: feedback on style can appear pedantic/tedious at first (and maybe forever), but is an important component of long term code maintenance.  There are tools to help accomplish PEP8 compliance.  Note: my initial commit may not have been compliant!

Even in its current form, the changes you made deserve some changes to the tests.  This is something we can discuss because that might not be obvious.  If/when you make a separate function, you’ll want to write new test(s) for that function.

dagmc_stats.py
```diff
@@ -30,24 +29,45 @@ def get_dagmc_tags(my_core):
...
+    meshset : a MOAB meshset to query for the ranges of entities
+    entity_ranges : a list of valid pyMOAB types to be retrieved
+    dagmc_tags : a dictionary of relevant tags. 
+    outputs
+    -------
+    entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
+    """
+    
+    
+    entity_ranges['Volumes'] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [3])
```
gonuke: `[3]` is a form of "magic number" - better to assign to a variable

dagmc_stats.py
```diff
@@ -30,24 +29,45 @@ def get_dagmc_tags(my_core):
...
+    entity_ranges['Surfaces'] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [2])
```
gonuke: This repeated line is begging for a loop:

```
for geom_dim, geom_name in enumerate(['Nodes', 'Curves', 'Surfaces', 'Volumes']):
    entity_ranges[geom_name] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [geom_dim])
```

gonuke: (we haven't really discussed Nodes, but a version of this is possible without)

generate_stats.py
```diff
@@ -32,13 +31,10 @@ def main():
 
     # get tags
     dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
-    
     # get Ranges of various entities
-    entity_types = [types.MBVERTEX, types.MBTRI]
-    entity_ranges = dagmc_stats.get_entity_ranges(my_core, all_meshset, entity_types)
-
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+    entity_ranges = dagmc_stats.get_entity_ranges(my_core, all_meshset, entity_types, dagmc_tags)
```
gonuke: need to update this to match new function name & add call to new function

dagmc_stats.py
```diff
@@ -1,25 +1,24 @@
 # set the path to find the current installation of pyMOAB
 import sys
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
+import numpy as np
```
gonuke: `numpy` isn’t used here - it’s best to avoid introducing unnecessary dependencies

dagmc_stats.py
```diff
@@ -1,25 +1,24 @@
...
-
 def get_dagmc_tags(my_core):
     """
     Get a dictionary with the important tags for DAGMC geometries
-
+    
     inputs
     ------
     my_core : a MOAB Core instance
-
```
gonuke: Better to keep this blank line for clarity/readability

dagmc_stats.py
```diff
@@ -30,24 +29,45 @@ def get_dagmc_tags(my_core):
     return dagmc_tags
 
 
-def get_entity_ranges(my_core, meshset, entity_types):
+def get_native_ranges(my_core, meshset, entity_types):
     """
-    Get a dictionary with MOAB ranges for each of the requested entity types
-
+    Get a dictionary with MOAB ranges for each of the native entity types
```
```suggestion
    Get a dictionary with MOAB ranges for each of the requested native entity types
```

dagmc_stats.py
```diff
@@ -30,24 +29,45 @@ def get_dagmc_tags(my_core):
...
+    """
+    Get a dictionary with MOAB ranges for each of the non-native entity types
+    
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    meshset : a MOAB meshset to query for the ranges of entities
+    entity_ranges : a list of valid pyMOAB types to be retrieved
+    dagmc_tags : a dictionary of relevant tags. 
+    outputs
```
```suggestion

    outputs
```

generate_stats.py
```diff
@@ -17,10 +17,9 @@ def report_stats(entity_ranges):
     ------
     entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
     """
-
     for entity_type, eh_range in entity_ranges.items():
         print("There are {} entities of type {}.".format(eh_range.size(),entity_type))
-
+    print('Type 0: Vertices \nType 2: Triangles \nType 11: EntitySets')
```
gonuke:
```suggestion
    print('Type 0: Vertices \nType 2: Triangles \nType 11: EntitySets')
    
```

dagmc_stats.py
```diff
@@ -3,23 +3,21 @@
...
-
     outputs
     -------
     dagmc_tags : a dictionary of relevant tags
     """
 
     dagmc_tags = {}
     dagmc_tags['geom_dim'] = my_core.tag_get_handle('GEOM_DIMENSION', size = 1, tag_type = types.MB_TYPE_INTEGER, #creates tag for the geo-
-                                                 storage_type = types.MB_TAG_SPARSE, create_if_missing = True)# metric dimension
+                                                 storage_type = types.MB_TAG_SPARSE, create_if_missing = True)# geometric dimension
```
bam241: Those line look very long. 
PEP8 (our reference style guide for python development) suggest 80 char per line. (there is tools allowing to auto format code to pep8 requirement (python package called `autopep8`). it is an easy way to clean the style of your the code

generate_stats.py
```diff
@@ -32,13 +34,11 @@ def main():
...
-    report_stats(entity_ranges)
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+    native_ranges = dagmc_stats.get_native_ranges(my_core, all_meshset, entity_types)
+    entityset_ranges = dagmc_stats.get_entityset_ranges(my_core, all_meshset, dagmc_tags['geom_dim'])
+    report_stats(native_ranges, entityset_ranges)
     
 if __name__ == "__main__":
-    main()
-    
+    main()
```
bam241: I would suggest to put back a blank line before the end of file. 
It is not exactly a pep8 requirement I believe (pep8 only requires blank line around class and method definitions....)

dagmc_stats.py
bam241: you can find docs on pep8 [[here](https://www.python.org/dev/peps/pep-0008/)]

dagmc_stats.py
```diff
@@ -3,23 +3,21 @@
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
```
bam241: this should be change at some point in the future....

generate_stats.py
bam241: you can find docs on pep8 [[here](https://www.python.org/dev/peps/pep-0008/)]

dagmc_stats.py
gonuke: This is a consequence of the current development environment....

PR #18, opened 2019-07-15
Tests Added for statistics functions

ep09772:
I added a few tests to the `nose` file to check that my stats functions for the dictionary were correct

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -3,15 +3,14 @@ import sys
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
 from pymoab import core, types
 from pymoab.rng import Range
-
+import numpy as np
+import matplotlib as plt
 def get_dagmc_tags(my_core):
     """
     Get a dictionary with the important tags for DAGMC geometries
-
     inputs
     ------
     my_core : a MOAB Core instance
-
     outputs
     -------
     dagmc_tags : a dictionary of relevant tags
@@ -30,16 +29,14 @@ def get_dagmc_tags(my_core):
     return dagmc_tags
 
 
-def get_entity_ranges(my_core, meshset, entity_types):
+def get_entity_ranges(my_core, meshset, entity_types, dagmc_tags):
     """
     Get a dictionary with MOAB ranges for each of the requested entity types
-
     inputs
     ------
     my_core : a MOAB Core instance
     meshset : a MOAB meshset to query for the ranges of entities
     entity_types : a list of valid pyMOAB types to be retrieved
-
     outputs
     -------
     entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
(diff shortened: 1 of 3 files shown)


test_dagmc_stats.py
```diff
@@ -7,17 +7,74 @@
...
+    values_list = list()
+    for val, count in test_set.items():
+        for count in range(count):
+            values_list.append(val)
+
+    median_computed = statistics.median(values_list)
+    median = dagmc_stats.find_median(test_set)
+    assert(median == median_manual == median_computed)
+    
+def test_median():
```
gonuke:
```suggestion
def test_mean():
```

test_dagmc_stats.py
```diff
@@ -7,17 +7,74 @@
 
 # import the new module that defines each of the functions
 import dagmc_stats
-
+from collections import OrderedDict
+import statistics
```
gonuke: `numpy` May have what you need here

test_dagmc_stats.py
```diff
@@ -7,17 +7,74 @@
...
+
+    median_computed = statistics.median(values_list)
+    median = dagmc_stats.find_median(test_set)
+    assert(median == median_manual == median_computed)
+    
+def test_median():
+    my_mean = dagmc_stats.find_mean(test_set)
+    values_list = list()
+    for val, count in test_set.items():
+        for count in range(count):
```
gonuke: Instead of this loop:

`values_list.append([val] * count)`

(Also... bad practice to use `count` as the loop variable when it has another meaning here)

ep09772: I honestly looked up another person's solution to the problem that is published online to see if our results matched. I don't know if that is bad practice or if I need to cite that somewhere, so if it is a problem I'll change it.

test_dagmc_stats.py
```diff
@@ -7,17 +7,74 @@
...
 def test_get_tags():
 
     dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
     assert(len(dagmc_tags) == 3)
 
+def test_get_median():
+
+    # Sort the dictionary
+    values_sorted = OrderedDict(sorted(test_set.items(), key=lambda t: t[0]))
+    index = sum(values_sorted.values())/2
```
bam241: Should not median calculation based on the number of occurrences in the set instead of their values ?

In a sorted set of values:
- if odd number of occurrences: median is the middle one:
        `(len(values_sorted)-1)/2`
- if even number of occurrences: median is the mean of the 2 middle values:
        `(len(values_sorted)/2-1` & `(len(values_sorted)/2+1`)

dagmc_stats.py
```diff
@@ -3,15 +3,13 @@
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
```
bam241: this should get change at some point.

dagmc_stats.py
```diff
@@ -3,15 +3,13 @@
...
 from pymoab import core, types
 from pymoab.rng import Range
-
+import numpy as np
 def get_dagmc_tags(my_core):
```
bam241: PEP8: 
-missing blank line, 
-import should be std lib first then custom

PR #19, opened 2019-07-16
Returns intermediate level statistics

ep09772:
#14 Resolved

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -3,15 +3,14 @@ import sys
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
 from pymoab import core, types
 from pymoab.rng import Range
-
+import numpy as np
+import matplotlib as plt
 def get_dagmc_tags(my_core):
     """
     Get a dictionary with the important tags for DAGMC geometries
-
     inputs
     ------
     my_core : a MOAB Core instance
-
     outputs
     -------
     dagmc_tags : a dictionary of relevant tags
@@ -30,16 +29,14 @@ def get_dagmc_tags(my_core):
     return dagmc_tags
 
 
-def get_entity_ranges(my_core, meshset, entity_types):
+def get_entity_ranges(my_core, meshset, entity_types, dagmc_tags):
     """
     Get a dictionary with MOAB ranges for each of the requested entity types
-
     inputs
     ------
     my_core : a MOAB Core instance
     meshset : a MOAB meshset to query for the ranges of entities
     entity_types : a list of valid pyMOAB types to be retrieved
-
     outputs
     -------
     entity_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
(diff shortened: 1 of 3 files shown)


dagmc_stats.py
```diff
@@ -3,15 +3,13 @@
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
 from pymoab import core, types
 from pymoab.rng import Range
-
+import numpy as np
```
zxkjack123: Suggestion: import `numpy` before `pymoab`.

dagmc_stats.py
```diff
@@ -30,16 +28,14 @@ def get_dagmc_tags(my_core):
     return dagmc_tags
 
 
-def get_entity_ranges(my_core, meshset, entity_types):
+def get_entity_ranges(my_core, meshset, entity_types, dagmc_tags):
```
zxkjack123: Suggestion: Add default value for parameter `dagmc_tags` for better compatibility.

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    Get the number of surfaces that each volume in a given file contains
+    
+    inputs
+    ------
+    my_core : a MOAB core instance
+    entity_ranges : a dictionary of the ranges of each tag in a file
+    
+    outputs
+    -------
+    pdf : a histogram that displays the number of surfaces each volume has (matplotlib)
```
zxkjack123:
```suggestion
    pdf : a histogram that displays the number of surfaces of each volume (matplotlib)
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    vert_list : a list of all vertices in a file
+    
+    outputs:
+    --------
+    sorted_dict: a dictionary that has information about triangles per vertex, sorted by its keys.
+    """
+    vertices_and_frequencies = {} 
+    #makes dictionary with each vertex and the number of triangles on it
+    for vert in vert_list:
+        number_of_vertex = vert
```
zxkjack123: Why not just use `vert` here? As the `number_of_vertex` is not used in other place.

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+        try:
+            number_of_frequencies[vertices_and_frequencies[vertex]] += 1
+        except KeyError:
+            number_of_frequencies[vertices_and_frequencies[vertex]] = 1
+    sorted_dict = {}
+    #sorts the dictionary
+    for frequency in sorted(number_of_frequencies.keys()):
+        sorted_dict[frequency] = number_of_frequencies[frequency]
+    return(sorted_dict)
+def find_median(sorted_dict):
```
zxkjack123:
```suggestion


def find_median(sorted_dict):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    for volumeset in entity_ranges['Volumes']:
+        freqs.append(my_core.get_child_meshsets(volumeset).size())
+    stats = {}
+    stats['minimum'] = min(freqs)
+    stats['maximum'] = max(freqs)
+    stats['median'] = np.median(freqs)
+    stats['average'] = np.average(freqs)
+    return stats
+
+def make_verts_dictionary(vert_list):
```
zxkjack123:
```suggestion

def make_verts_dictionary(vert_list):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
     for entity_type in entity_types:
         entity_ranges[entity_type] = my_core.get_entities_by_type(meshset, entity_type) 
-
+        if entity_type == 11:
+            entity_ranges['Volumes'] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [3])
+            entity_ranges['Surfaces'] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [2])
+            entity_ranges['Curves'] = my_core.get_entities_by_type_and_tag(meshset, types.MBENTITYSET, dagmc_tags['geom_dim'], [1])
     return entity_ranges
 
+def get_surfaces_per_volume(my_core, entity_ranges):
```
zxkjack123:
```suggestion

def get_surfaces_per_volume(my_core, entity_ranges):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    
+    inputs
+    ------
+    sorted_dict : a dictionary sorted by its keys
+    
+    outputs
+    -------
+    mean : the median of the dataset
+    
+    example dictionary:
```
zxkjack123: This example dictionary seems to be a input parameter, suggest to put it in the block of `inputs`.

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    {3:2, 4:5, 6:3} = [3,3,4,4,4,4,4,6,6,6]
+    """
+    #finds the number of data points
```
zxkjack123: Suggest to use `len(sorted_dict)` to get the number of data points.

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    if sum(list(sorted_dict.values())[index:]) != sum(list(sorted_dict.values())[:index]):
+        if sum(list(sorted_dict.values())[index:]) > sum(list(sorted_dict.values())[:index]):
+            median = list(sorted_dict.keys())[index]
+        else:
+            median = list(sorted_dict.keys())[index-1]
+    else:
+        median = (list(sorted_dict.keys())[index-1] + list(sorted_dict.keys())[index]) / 2
+    return(median)
+
+def find_mean(sorted_dict):
```
zxkjack123:
```suggestion

def find_mean(sorted_dict):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    
+    example dictionary:
+    {3:2, 4:5, 6:3} = [3,3,4,4,4,4,4,6,6,6]
+    """
+    total = 0
+    for key, value in sorted_dict.items():
+        total += key*value
+    mean = total / sum(sorted_dict.values())
+    return(mean)
+def get_triangles_per_vertex(my_core, all_meshset):
```
zxkjack123:
```suggestion


def get_triangles_per_vertex(my_core, all_meshset):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    """
+    gets stats for the number of trianges touching each vertex in a file
+    inputs
```
zxkjack123:
```suggestion

    inputs
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    #makes a dictionary of all statistics
+    triangles_per_vertex_stats = {}
+    triangles_per_vertex_stats['minimum'] = min(sorted_dict.keys())
+    triangles_per_vertex_stats['maximum'] = max(sorted_dict.keys())
+    median = find_median(sorted_dict)
+    triangles_per_vertex_stats['median'] = median
+    mean = find_mean(sorted_dict)
+    triangles_per_vertex_stats['average'] = mean
+    return triangles_per_vertex_stats
+def get_triangles_per_surface(my_core, entity_ranges):
```
zxkjack123:
```suggestion


def get_triangles_per_surface(my_core, entity_ranges):
```

dagmc_stats.py
```diff
@@ -48,6 +44,186 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    '''
```
zxkjack123: Use `"""` here.

dagmc_stats.py
ep09772: If you look at the example dictionary, the length of the dictionary isn't what I'm after. Instead, I want to find the length of the subsequent list. For the example, `len(example_dict) = 3`, but the value I want is 2+5+3 = 10 (or `example_dict[3] + example_dict[4] +example_dict[5]`).

dagmc_stats.py
ep09772: The default value that would be put there is generated from the previous function; how could I transfer that from the other function to be a parameter of this function?

dagmc_stats.py
zxkjack123: Got it. Then I suggest to try:
```python
length = sum(sorted_dict.values())
```
Eg:
```python
>>> d = {3:2, 4:5, 6:3}
>>> sum(d.values())
10
```

dagmc_stats.py
zxkjack123: The `dagmc_tags` are used when and only when `entity_type == 11`. Therefore, it is an optional parameter. For most cases, you don't need to pass in this parameter when you call this function. Add a default value could make it cleaner.

You can just set it to be `None`. And check the value before you use it:

ep09772: Thanks, will implement this

PR #23, opened 2019-07-18
Returns Intermediate Level Statistics

ep09772:


ep09772:
I cleaned up the pull requests from earlier, but tried to keep one for each step I took. Let me know if I should reopen more, or if I should delete still more.

gonuke:
At some point we should discuss a generic design for compiling statistics.  It’s probably best to make a big list (or maybe better a `numpy` array) of all the values and then use python tools to perform the stats and plots.

PR #26, opened 2019-07-23
Surfaces per volume

ep09772:
Functions, statistics, tests, and reporting of surface per volume


dagmc_stats.py
```diff
@@ -42,12 +44,56 @@ def get_entity_ranges(my_core, meshset, entity_types):
...
+    my_core : a MOAB core instance
+    entity_ranges : a dictionary of the entityset ranges of each tag in a file
+
+    outputs
+    -------
+    freqs : a list of the number of surfaces each volume in the file contains
+
+    """
 
+    freqs = []
```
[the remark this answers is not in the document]
gonuke: Also - this data is probably not actually frequencies, so it may be best to use a different name.  Accurate variable naming is good for improving long term maintainability.

generate_stats.py
```diff
@@ -1,44 +1,97 @@
...
+#This file is the script that users will actually run to generate the full set of statistics for a file
 
 # set the path to find the current installation of pyMOAB
 import sys
 sys.path.append('/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
 from pymoab import core, types
 from pymoab.rng import Range
-
 # import the new module that defines each of the functions
 import dagmc_stats
```
gonuke: Import your own custom module last

generate_stats.py
```diff
@@ -1,44 +1,97 @@
...
+import argparse
+import numpy as np
+
+
+
 
-def report_stats(entity_ranges):
+
```
gonuke:
```suggestion
```
Only need 1 blank link

generate_stats.py
```diff
@@ -1,44 +1,97 @@
...
+ 
+    """
+    stats = {}
+    dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+    native_ranges = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)     # get Ranges of various entities
+    entityset_ranges = dagmc_stats.get_entityset_ranges(my_core, root_set, dagmc_tags['geom_dim'])
+    s_p_v_data = dagmc_stats.get_surfaces_per_volume(my_core, entityset_ranges)
+    s_p_v_stats = get_stats(s_p_v_data)
+    stats['S_P_V'] = s_p_v_stats
```
gonuke: we should probably store the data as well as the stats for use in visualization, etc

generate_stats_filename.py
```diff
@@ -0,0 +1,73 @@
+#This file is the script that users will actually run to generate the full set of statistics for a file
```
gonuke: I don't understand why this file is here?

ep09772: I was playing around with it before I implemented argparse, I don't know why it's still here.

test_dagmc_stats.py
```diff
@@ -53,3 +53,12 @@ def test_get_entityset_ranges():
     volume_range = my_core.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [3])
     assert(volume_range == entityset_ranges['Volumes'])
     
+
+def test_get_surfaces_per_volume():
```
gonuke: In this model, we know that there are 3 volumes: 2 cubes and a sphere, and that the cubes have 6 surfaces per volume and the sphere has 1.  We should be able to test the actual values of this data, not just the length of the list.

generate_stats.py
```diff
@@ -53,8 +71,12 @@ def collect_statistics(my_core, root_set):
     stats = {}
     dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
     entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
-    stats['Native Ranges'] = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)     # get Ranges of various entities
-    stats['EntitySet Ranges'] = dagmc_stats.get_entityset_ranges(my_core, root_set, dagmc_tags['geom_dim']) 
+    native_ranges = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)     # get Ranges of various entities
```
gonuke: This is mostly a matter of preference, but for readability, I'd divide this block up with carefully placed blank lines.  Think of it like dividing prose up into paragraphs.  Some lines belong together because they are related but in blocks they can be separated somewhat.  For example, you could have a "block" for each of the following:
* getting the tags
* getting the native ranges
* getting the geometry set ranges
* getting and processing the s_p_v data

dagmc_stats.py
```diff
@@ -72,3 +75,24 @@ def get_entityset_ranges(my_core, meshset, geom_dim):
...
+    my_core : a MOAB core instance
+    entity_ranges : a dictionary of the entityset ranges of each tag in a file
+
+    outputs
+    -------
+    freqs : a list of the number of surfaces each volume in the file contains
+
+    """
+
+    s_p_v = []
```
gonuke: why not declare it as an `np.array` right away?
```suggestion
    s_p_v = np.array()
```

ep09772: using `np.array()` with no arguments raises an error

ep09772: But I was able to work around it

generate_stats.py
```diff
@@ -51,11 +63,19 @@ def collect_statistics(my_core, root_set):
...
+    
     entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
-    stats['Native Ranges'] = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)     # get Ranges of various entities
-    stats['EntitySet Ranges'] = dagmc_stats.get_entityset_ranges(my_core, root_set, dagmc_tags['geom_dim']) 
-    return stats
+    native_ranges = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)     # get Ranges of various entities
+    
+    entityset_ranges = dagmc_stats.get_entityset_ranges(my_core, root_set, dagmc_tags['geom_dim'])
+    
+    data['S_P_V_data'] = dagmc_stats.get_surfaces_per_volume(my_core, entityset_ranges)
```
gonuke: Good idea to separate this into two dictionaries!  It might make life easier down the road if you use the same key for both, possibly even using a variable to store that key for consistency:

```
s_p_v_key = ‘S_P_V’
data[s_p_v_key] = ....
stats[s_p_v_key] = ...
```

ep09772:
I think this PR should be good to merge

Issue #38, opened 2019-08-09
triangle aspect ratio distribution


kkiesling:
Create a PDF of the aspect ratio of triangles in the geometry. Triangle aspect ratio can be calculated according to [this equation](https://codegolf.stackexchange.com/questions/101234/evaluate-the-aspect-ratio-of-a-triangle). It would be good to have the option to get this distribution for the entire geometry, a specific volume, or a specific surface.

PR #47, opened 2019-08-19
Triangle aspect

ep09772:
This PR introduces the obtaining and reporting of the triangle aspect ratio. This is a relatively raw PR, and there are some enhancement features that will need to be added later (ie choosing what meshset to get the stats from).

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -1,24 +1,28 @@
 # set the path to find the current installation of pyMOAB
 import sys
 import numpy as np
+import math
 sys.path.append(
     '/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
 from pymoab.rng import Range
 from pymoab import core, types
 
+
 def get_dagmc_tags(my_core):
     """
     Get a dictionary with the important tags for DAGMC geometries
-
+    
     inputs
     ------
     my_core : a MOAB Core instance
+    
     outputs
     -------
     dagmc_tags : a dictionary of relevant tags
     """
 
     dagmc_tags = {}
+
     dagmc_tags['geom_dim'] = my_core.tag_get_handle('GEOM_DIMENSION', size=1, tag_type=types.MB_TYPE_INTEGER,   
                                                     storage_type=types.MB_TAG_SPARSE, create_if_missing=True)  # geometric dimension
 
@@ -26,6 +30,7 @@ def get_dagmc_tags(my_core):
                                                     storage_type=types.MB_TAG_SPARSE, create_if_missing=True)  # the category
 
     dagmc_tags['global_id'] = my_core.tag_get_handle('GLOBAL_ID', size=1, tag_type=types.MB_TYPE_INTEGER,  
+
                                                      storage_type=types.MB_TAG_SPARSE, create_if_missing=True)  # id
 
     return dagmc_tags
@@ -34,16 +39,17 @@ def get_dagmc_tags(my_core):
 def get_native_ranges(my_core, meshset, entity_types):
     """
     Get a dictionary with MOAB ranges for each of the requested entity types
-
+    
     inputs
     ------
     my_core : a MOAB Core instance
     meshset : a MOAB meshset to query for the ranges of entities
     entity_types : a list of valid pyMOAB types to be retrieved
-
+    
     outputs
     -------
-    native_ranges : a dictionary with one entry for each entity type that is a Range of handles to that type
+    native_ranges : a dictionary with one entry for each entity type that is a
+                    Range of handles to that type
     """
 
     native_ranges = {}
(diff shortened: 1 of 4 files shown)


ep09772:
Will resolve #38

gonuke:
I think this PR needs to be rebased against master - it appears to include some changes that have already been merged.

dagmc_stats.py
```diff
@@ -55,44 +61,147 @@ def get_native_ranges(my_core, meshset, entity_types):
...
+    tris = my_core.get_entities_by_type(meshset, types.MBTRI)
+    
+    for triangle in tris:
+        side_lengths = []
+        s = 0
+        coord_list = []
+        
+        verts = list(my_core.get_adjacencies(triangle, 0))
+        
+        for side in range(3):
```
gonuke: you should be able to get the coordinates of all three vertices in one call

dagmc_stats.py
```diff
@@ -55,44 +61,147 @@ def get_native_ranges(my_core, meshset, entity_types):
...
+            coords = list(my_core.get_coords(verts[side]))
+            coord_list.append(coords)
+            
+        coord_array = np.array(coord_list)
+        
+        for length in range(3):    
+            sum_squares = 0
+            
+            for vert in range(3):
```
gonuke: With the right data structure, you should be able to do this as a numpy vector operation.

See this, for example: https://stackoverflow.com/questions/1401712/how-can-the-euclidean-distance-be-calculated-with-numpy

dagmc_stats.py
ep09772: From what I see about getting the Euclidean distance between two coordinates, it may be best for the xyz data to be in separate arrays, but if you get all coordinates in one call, they would be returned in a 1-D array.

dagmc_stats.py
ep09772: This works very nicely with a list of arrays

dagmc_stats.py
ep09772: I think a list of 1-D arrays would work better for getting the distance between the points

dagmc_stats.py
```diff
@@ -151,3 +152,47 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+    tris = my_core.get_entities_by_type(meshset, types.MBTRI)
+    
+    for triangle in tris:
+        side_lengths = []
+        s = 0
+        coord_list = []
+        
+        verts = list(my_core.get_adjacencies(triangle, 0))
+        
+        for side in range(3):
```
```suggestion
        for vert in verts:
```
(and corresponding change in the next line)

dagmc_stats.py
```diff
@@ -151,3 +152,47 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+            coords = my_core.get_coords(verts[side])
+            coord_list.append(coords)
+        
+        for length in range(3):    
+            side_lengths.append(np.linalg.norm(coord_list[length]-coord_list[indexes[length]]))
+            
+        s = .5*(sum(side_lengths))
+        top = np.prod(side_lengths)
+        bottom = 8*(s-side_lengths[0])*(s-side_lengths[1])*(s-side_lengths[2])
```
```suggestion
        bottom = 8*np.prod(s-side_lengths)
```
`numpy` will do this scalar - vector correctly

dagmc_stats.py
```diff
@@ -151,3 +152,47 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+        s = 0
+        coord_list = []
+        
+        verts = list(my_core.get_adjacencies(triangle, 0))
+        
+        for side in range(3):
+            coords = my_core.get_coords(verts[side])
+            coord_list.append(coords)
+        
+        for length in range(3):
```
```suggestion
        for side in range(3):    
```

A more accurate variable name, with corresponding change(s) in next line

dagmc_stats.py
```diff
@@ -151,3 +152,47 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+        coord_list = []
+        
+        verts = list(my_core.get_adjacencies(triangle, 0))
+        
+        for side in range(3):
+            coords = my_core.get_coords(verts[side])
+            coord_list.append(coords)
+        
+        for length in range(3):    
+            side_lengths.append(np.linalg.norm(coord_list[length]-coord_list[indexes[length]]))
```
gonuke: I see what you are doing here, but I wonder if it would be clearer if you had defined (above, instead of `coord_list`):

```
first_coord = [0, 1, 2]
second_coord = [1, 2, 0]
```

and then wrote this as `first_coord[side] - second_coord[side]`?

dagmc_stats.py
```diff
@@ -1,6 +1,7 @@
 # set the path to find the current installation of pyMOAB
 import sys
 import numpy as np
+import math
```
gonuke: are you still using this, or just using `numpy` methods?

dagmc_stats.py
ep09772: I don't quite understand how you recommend changing the above definitions. I do believe that if we are able to write it as `first_coord[side] - second_coord[side]` that it would be clearer, but I don't know how to implement them

ep09772: That being said, I did come up with something that makes it easier to read, but may not clarify exactly what is being done.

dagmc_stats.py
```diff
@@ -182,16 +180,16 @@ def get_triangle_aspect_ratio(my_core, meshset):
...
-        for side in range(3):
-            coords = my_core.get_coords(verts[side])
+        for vert in verts:
+            coords = my_core.get_coords(vert)
             coord_list.append(coords)
         
-        for length in range(3):    
-            side_lengths.append(np.linalg.norm(coord_list[length]-coord_list[indexes[length]]))
+        for side in range(3):    
+            side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
```
gonuke: definitely more readable!!  Maybe a little too clever :smile: 

This probably deserves a comment like:
```
# triangle sides go between pairs of vertices with IDs: (0,1), (1,2), (2,0)
# if interpreted as indices into a list of length 3, these can be accomplished with (0,-2), (1,-1), (2,0), or (side, 2-side)
```

dagmc_stats.py
```diff
@@ -151,3 +151,46 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+        coord_list = []
+        
+        verts = list(my_core.get_adjacencies(triangle, 0))
+        
+        for vert in verts:
+            coords = my_core.get_coords(vert)
+            coord_list.append(coords)
+        
+        for side in range(3):    
+            side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
```
gonuke: definitely more readable!!  Maybe a little too clever :smile: 

This probably deserves a comment like:
```
# triangle sides go between pairs of vertices with IDs: (0,1), (1,2), (2,0)
# if interpreted as indices into a list of length 3, these can be accomplished with (0,-2), (1,-1), (2,0), or (side, 2-side)
```

gonuke: Note that we don't need comments in most places because good variable names and data structure choices make the code more obvious to follow.  I think this is probably a case where it is a reasonable approach, but not obvious to follow.

Issue #31, opened 2019-08-08
add argparse options for statistics


kkiesling:
Add argparse options to allow users to generate statistics for only the statistics needed. Default use can be to generate all statistics.

PR #49, opened 2019-08-21
Argparse options

ep09772:
Creates options for the user to implement on the command line. For example, the user can specify which statistics they specifically want to be displayed.


ep09772:
will resolve #31

Issue #37, opened 2019-08-08
Write entity specific output file


kkiesling:
I am not sure if this would actually be helpful, but for stats such as tris/vertex (maybe less so this one), tris/surf, and surfs/vol, writing an output file that lists the surf or volume ID or entity handle and the number of tris, surfs, etc. This could be an optional output file but might be helpful if the user wants to know the data for a specific surface or volume. I think it would be a little too much to do the tris/vertex, but tris/surf and surfs/vol could be doable.

Example output:
```
Surface-ID   Triangles/surface
    1          12
    2          45
    3          5
...    ...
```

Another way to get this information would be to make it an argparse option to specify a specific volume or surface ID and then print to screen the # triangles, surfaces, etc.

kkiesling:
I am also trying to look at PRs #26, #27, and #28 to see if the information is essentially already reported out.

PR #50, opened 2019-08-26
Entity Specific Stats

ep09772:
Prints the stats for each entityhandle relevant to a statistical area. See #37


gonuke:
I think this does what we need.  I can imagine a few new issues that we've discussed that would be good to write up, but otherwise it't good to go.

PR #54, opened 2019-10-25
Funtion for triangle area calculation in dagmc_stats.py file and its partial test

yqin43:
## Description
method for triangle area calculation in dagmc_stats.py file and its partial test

## Motivation and Context
calculate the areas of triangles

## Changes
In dagmc_stats.py: added function for triangle area calculation(get_area_triangle)
In generate_stats.py: added code to call the get_area_triangle function
In test_dagmc_stats.py: added partial test to test the get_area_triangle function. Tests if the number of areas generated is equal to the number of triangles.

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -154,9 +154,87 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
     return s_p_v
 
 
+def get_tris(my_core, meshset, geom_dim):
+    """
+    get triangles
+    inputs
+    ------
+    my_core : a MOAB core instance
+    entity_ranges : a dictionary of the entityset ranges of each tag in a file
+    geom_dim : a MOAB Tag that holds the dimension of an entity.
+
+    outputs
+    -------
+    tris : (list)triangles
+    """
+    if my_core.tag_get_data(geom_dim, meshset)[0][0] == 3:
+        entities = my_core.create_meshset()
+        for surface in my_core.get_child_meshsets(meshset):
+            my_core.add_entities(entities, my_core.get_entities_by_type(surface, types.MBTRI))
+        tris = my_core.get_entities_by_type(entities, types.MBTRI)
+    else:
+        tris = my_core.get_entities_by_type(meshset, types.MBTRI)
+    return tris
+
+
+def get_tri_side_length(my_core, tri):
+    """
+    get side lengths of triangle
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    tri : triangle
+
+    outputs
+    -------
+    side_lengths : (list)side lengths of triangle
+    """
+    side_lengths = []
+    s = 0
+    coord_list = []
+
+    verts = list(my_core.get_adjacencies(tri, 0))
+
+    for vert in verts:
+        coords = my_core.get_coords(vert)
+        coord_list.append(coords)
+
+    for side in range(3):
+        side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
+        # The indices of coord_list includes the "-2" because this way each side will be matched up with both
+        # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
+    return side_lengths
+
+
 def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
     """
     Gets the triangle aspect ratio (according to the equation: (abc)/(8(s-a)(s-b)(s-c)), where s = .5(a+b+c).)
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    meshset : a meshset containing a certain part of the mesh
+    geom_dim : a MOAB Tag that holds the dimension of an entity.
+
+    outputs
+    -------
+    t_a_r : (list) the triangle aspect ratios for all triangles in the meshset
+    """
+    tris = get_tris(my_core, meshset, geom_dim)
+    t_a_r = []
+
+    for tri in tris:
+        side_lengths = get_tri_side_length(my_core, tri)
+        s = .5*(sum(side_lengths))
+        top = np.prod(side_lengths)
+        bottom = 8*np.prod(s-side_lengths)
+        t_a_r.append(top/bottom)
+
+    return t_a_r
+
+def get_area_triangle(my_core, meshset, geom_dim):	
+    """
+    Gets the triangle area (according to the equation: sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2)
     
     inputs
     ------
@@ -166,75 +244,14 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
     
     outputs
     -------
-    t_a_r : (list) the triangle aspect ratios for all triangles in the meshset
-    """
-
-    if my_core.tag_get_data(geom_dim, meshset)[0][0] == 3:
-        entities = my_core.create_meshset()
-        for surface in my_core.get_child_meshsets(meshset):
-            my_core.add_entities(entities, my_core.get_entities_by_type(surface, types.MBTRI))
-        tris = my_core.get_entities_by_type(entities, types.MBTRI)
-    else:
-        tris = my_core.get_entities_by_type(meshset, types.MBTRI)
-
-    t_a_r = []
-
-    for tri in tris:
-        side_lengths = []
-        s = 0
-        coord_list = []
-
-        verts = list(my_core.get_adjacencies(triangle, 0))
-
-        for vert in verts:
-            coords = my_core.get_coords(vert)
-            coord_list.append(coords)
-
-        for side in range(3):    
-            side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
-            # The indices of coord_list includes the "-2" because this way each side will be matched up with both
-            # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
-      
-        s = .5*(sum(side_lengths))
-        top = np.prod(side_lengths)
-        bottom = 8*np.prod(s-side_lengths)
-        t_a_r.append(top/bottom)
-
-    return t_a_r
-
-def get_area_triangle(my_core, meshset):	
-    """
-    Gets the triangle area (according to the equation: sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2)
-    
-    inputs
-    ------
-    my_core : a MOAB Core instance
-    meshset : a meshset containing a certain part of the mesh
-    
-    outputs
-    -------
     area : (list) the triangle areas for all triangles in the meshset
     """
     
     area = []
-    tris = my_core.get_entities_by_type(meshset, types.MBTRI)
+    tris = get_tris(my_core, meshset, geom_dim)
 
     for tri in tris:
-        side_lengths = []
-        s = 0
-        coord_list = []
-
-        verts = list(my_core.get_adjacencies(triangle, 0))
-
-        for vert in verts:
-            coords = my_core.get_coords(vert)
-            coord_list.append(coords)
-
-        for side in range(3):
-            side_lengths.append(np.linalg.norm(coord_list[side] - coord_list[side - 2]))
-            # The indices of coord_list includes the "-2" because this way each side will be matched up with both
-            # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
-
+        side_lengths = get_tri_side_length(my_core, tri)
         # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
         s = sum(side_lengths)/2
         s = np.sqrt(s * np.prod(s - side_lengths))
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -1,7 +1,6 @@
 # set the path to find the current installation of pyMOAB
 import sys
 import numpy as np
-import math
 
 sys.path.append(
     '/opt/tljh/user/lib/moab/lib/python3.6/site-packages/pymoab-5.1.0-py3.6-linux-x86_64.egg')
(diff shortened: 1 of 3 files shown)


dagmc_stats.py
```diff
@@ -201,3 +203,43 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
...
+        for vert in verts:
+            coords = my_core.get_coords(vert)
+            coord_list.append(coords)
+
+        for side in range(3):
+            side_lengths.append(np.linalg.norm(coord_list[side] - coord_list[side - 2]))
+            # The indices of coord_list includes the "-2" because this way each side will be matched up with both
+            # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
+
+        s = math.sqrt(sum(side_lengths) * np.prod(sum(side_lengths) - side_lengths))
```
gonuke: Recommend using the `sqrt` from the `numpy` package.

dagmc_stats.py
```diff
@@ -201,3 +203,43 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
         t_a_r.append(top/bottom)
 
     return t_a_r
+
+def get_area_triangle(my_core, meshset):
+	
+	"""
+    Gets the triangle area (according to the equation: sqrt(s(s - a)(s - b)(s - c)), where s = a + b + c)
```
gonuke: I'm not sure this formula is correct.... do you have a reference?

Consider a 3/4/5 triangle...  s = 12, A = sqrt(12 * 9 * 8 * 7 ) = 77.7689.  

However, using the more conventional formula gives: A = b*h/2 = 3 * 4 /2 = 6

I think s = 1/2 * (a+b+c)

https://www.mathopenref.com/heronsformula.html

generate_stats.py
```diff
@@ -197,17 +210,19 @@ def main():
...
     verbose = args.verbose
     tps_data = args.tps_data
     spv_data = args.spv_data
     display_options = {'NR':args.nr, 'ER':args.er, 'SPV':args.spv, 'TPV':args.tpv,
-                       'TPS':args.tps, 'TAR':args.tar, 'TPS_data':args.tps_data, 'SPV_data':args.spv_data} 
+                       'TPS':args.tps, 'TAR':args.tar, 'AT': args.at, 'TPS_data':args.tps_data, 'SPV_data':args.spv_data} 
     if not(True in display_options.values()):
         display_options = {'NR':True, 'ER':True, 'SPV':True, 'TPV':True, 'TPS':True,
-                           'TAR':True, 'TPS_data':False, 'SPV_data':False}
+                'TAR':True, 'AT':True, 'TPS_data':False, 'SPV_data':False}
```
gonuke: please indent this for readability

test_dagmc_stats.py
```diff
@@ -99,3 +99,11 @@ def test_get_surfaces_per_volume():
     for eh in range(known_volumes.size()):
         surfs = my_core.get_child_meshsets(known_volumes[eh]).size()
         assert(surfs == s_p_v_data[eh])
+
+def test_get_area_triangle():
```
gonuke: This matches the tests for previous functions, but we probably need to start improving the scope of testing.

For example, it would be wise to split out the area function to two pieces: one that gets the data from MOAB and another that uses that data to calculate the area.  Then it would be possible to test the function that calculates the area with a variety of data/cases that are independent of MOAB.

kkiesling: We did discuss this issue with the testing. The current geometry used in these tests has too many triangles and volumes to easily figure out which triangle is which and calculate the size of it by hand. I advised Yuehan to make this PR so we could review it first and then make a second PR with a more useable geometry for better testing (I have to make a better geometry file for her still). Given the question of whether the area calculation is correct, it might be worthwhile making sure it gets in with this PR as well.

dagmc_stats.py
```diff
@@ -201,3 +203,43 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
         t_a_r.append(top/bottom)
 
     return t_a_r
+
+def get_area_triangle(my_core, meshset):
```
kkiesling: This function has many similarities to the function above `get_triangle_aspect_ratio`. The only difference seems to be the formula used to calculate area or aspect ratio. The two could probably be combined into a single function (such as `get_triangle_stats` or similar) and then only calls functions that do the two different maths like `calc_aspect_ratio` and `calc_tri_area`.

```
def calc_aspect_ratio(...):
    tar = ...
    return tar

def calc_tri_area(...):
    area = ...
    return area

def get_triangle_stats(...):
     # do all the work one time to collect the list of verts and calculating a, b, c for the triangle
     if TAR:
         calc_aspect_ratio()
     if area:
         calc_tri_area()
    return ...
```
That's one example to consolidate these two functions. There may be a better/different way. But the ultimate goal should be to reduce redundant code.

dagmc_stats.py
kkiesling: I think this @gonuke is right, `s=1/2*(a+b+c)` (https://www.mathsisfun.com/geometry/herons-formula.html)

yqin43: Yes, I am using the Heron's formula but calculated the wrong s. I will fix it. Thank you!

test_dagmc_stats.py
gonuke: What do you think about separating out the method for calculating the area from the method for getting data from MOAB?

kkiesling: Yes, I think that is a good separation to make too. I made a comment (sort of) about this on `dagmc_stats.py`. The separation makes sense for the reason you state here and because of the repeated code between this function and the triangle_aspect_ratio function.

dagmc_stats.py
```diff
@@ -201,3 +201,43 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
...
+    
+    outputs
+    -------
+    area : (list) the triangle areas for all triangles in the meshset
+    """
+    
+    area = []
+    tris = my_core.get_entities_by_type(meshset, types.MBTRI)
+
+    for triangle in tris:
```
bam241: Not a big deal but I would use more consistent naming convention:
`verts`/`vert`
`tris`/`tri` or `triangles`/`triangle`

dagmc_stats.py
```diff
@@ -201,3 +201,43 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
...
+
+        for side in range(3):
+            side_lengths.append(np.linalg.norm(coord_list[side] - coord_list[side - 2]))
+            # The indices of coord_list includes the "-2" because this way each side will be matched up with both
+            # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
+
+        s = sum(side_lengths)/2
+        s = np.sqrt(s * np.prod(s - side_lengths))
+        area.append(s)
+        # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
```
bam241: This comment should be located where `s` is calculated (just before or at the end of the line) not after the append...

bam241: And it is already in the method doctoring, so might not be required here :)

yqin43:
Hi everyone,

I have updated the code and please have a look! Thank you!

I have edited my code based on the comments except combing two functions to one. I find that there are many other things to change in the generate_stats.py file corresponding to combing the two functions so I prefer not to combine the two functions as it is pretty complex. Thank you!

gonuke:
I don't think the suggestion was to combine two functions into one.  Rather, I think the suggestion is make 2 functions into 3. 

Perhaps an easy first step will be to write a function that calculates the area of a triangle given three coordinates.  Even if it wasn't used in any other function, it would dramatically help with testing in the current function.

kkiesling:
Thanks for making these changes @yqin43 - we'll talk in person about how we can refactor what we have according to my and @gonuke's comments.

dagmc_stats.py
```diff
@@ -153,51 +154,111 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
-    t_a_r = []
 
-    for triangle in tris:
-        side_lengths = []
-        s = 0
-        coord_list = []
+def get_tri_side_length(my_core, tri):
+    """
+    get side lengths of triangle
+    inputs
```
kkiesling: should have a blank line above this

dagmc_stats.py
```diff
@@ -153,51 +154,111 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+    ------
+    my_core : a MOAB Core instance
+    tri : triangle
```
kkiesling: should include data type. I assume this is an entity handle for an MOAB tri.

dagmc_stats.py
```diff
@@ -153,51 +154,111 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+
+    for tri in tris:
+        side_lengths = get_tri_side_length(my_core, tri)
         s = .5*(sum(side_lengths))
         top = np.prod(side_lengths)
         bottom = 8*np.prod(s-side_lengths)
         t_a_r.append(top/bottom)
 
     return t_a_r
+
```
kkiesling: PEP8 - add a second blank line here

generate_stats.py
```diff
@@ -163,7 +171,12 @@ def collect_statistics(my_core, root_set, tar_meshset, display_options):
         data[tar_key] = dagmc_stats.get_triangle_aspect_ratio(
                                     my_core, tar_meshset, dagmc_tags['geom_dim'])
         stats[tar_key] = get_stats(data[tar_key])
-        
+   
+    if display_options['AT']:
+        at_key = 'A_T'
```
kkiesling: Is there a reason these keys are different `AT` vs `A_T`?

dagmc_stats.py
```diff
@@ -153,51 +154,111 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+    """
+    tris = get_tris(my_core, meshset, geom_dim)
+    t_a_r = []
+
+    for tri in tris:
+        side_lengths = get_tri_side_length(my_core, tri)
         s = .5*(sum(side_lengths))
         top = np.prod(side_lengths)
         bottom = 8*np.prod(s-side_lengths)
         t_a_r.append(top/bottom)
```
kkiesling: These lines here could be moved into a function `calc_tri_aspect_ratio(side_lengths)` like @gonuke suggested. Then it could be called here and it would be easier to test.

@gonuke correct me if this is not what you're suggesting.

dagmc_stats.py
```diff
@@ -153,51 +154,111 @@ def get_surfaces_per_volume(my_core, entityset_ranges):
...
+    
+    area = []
+    tris = get_tris(my_core, meshset, geom_dim)
+
+    for tri in tris:
+        side_lengths = get_tri_side_length(my_core, tri)
+        # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
+        s = sum(side_lengths)/2
+        s = np.sqrt(s * np.prod(s - side_lengths))
+        area.append(s)
```
kkiesling: same as my above comment and what I think @gonuke is suggesting. These lines could be moved into their own function that uses `side_lengths` as input and returns `area` to be appended in the list.

yqin43:
Hi everyone, I made the edits and please have a look! Thank you!

kkiesling:
> Hi everyone, I made the edits and please have a look! Thank you!

@yqin43 I don't see any updates to making the `get_tri_area(side_lengths)` and `get_tri_aspect_ratio(side_lengths)` functions? Did you make those changes too?

kkiesling:
> > Hi everyone, I made the edits and please have a look! Thank you!
> 
> @yqin43 I don't see any updates to making the `get_tri_area(side_lengths)` and `get_tri_aspect_ratio(side_lengths)` functions? Did you make those changes too?

Never mind! I see where these changes were made. Sorry for the delay.

This looks good to me now! I'll let @gonuke approve before merging.

test_dagmc_stats.py
```diff
@@ -99,3 +90,33 @@ def test_get_surfaces_per_volume():
...
+def test_get_triangle_aspect_ratio():
+    """
+    Tests part of the get_triangle_aspect_ratio function
+    """
+    test_input2 = "single-cube.h5m"
+    my_core = core.Core()
+    my_core.load_file(test_input2)
+    root_set = my_core.get_root_set()
+
+    assert(abs(dagmc_stats.get_triangle_aspect_ratio(my_core, root_set) - 1000*np.sqrt(2)/400/(10-5*np.sqrt(2))) < 0.01)
```
gonuke: Doesn't this call to `get_triangle_aspect_ratio` return a list of many aspect ratios?  I'm guessing this single cube model has 12 triangles.  If so, I'm not sure exactly how to interpret this comparison.  It is perhaps coincidence that ll 12 triangles have the same aspect ratio?

Also, it is good practice to document a test in this form:
```
exp = 1000*np.sqrt(2)/400/(10-5*np.sqrt(2))) < 0.01
obs = dagmc_stats.get_triangle_aspect_ratio(my_core, root_set)
assertAlmostEqual(exp,obs)
```

Notes:
* `exp` refers to the expected result
* `obs` refers to the observed result
* `assertAlmostEqual` is a method in `nose` (I think) that allows you to handle floating point comparison that may not be exact.

Similar improvements could be made in the triangle area test below.

yqin43: I have edited the test file and please have a look again. Thank you!

For the single cube, it has 12 identical triangles with the same aspect ratio. Since model with various aspect ratio is hard to test, I choose to run test on the single cube to see if it computes the correct aspect ratio for the triangle.

Thanks for all the comments. I learned a lot from them.

test_dagmc_stats.py
```diff
@@ -99,3 +91,39 @@ def test_get_surfaces_per_volume():
...
+    test_input2 = "single-cube.h5m"
+    my_core = core.Core()
+    my_core.load_file(test_input2)
+    root_set = my_core.get_root_set()
+
+    exp = 1000*np.sqrt(2)/400/(10-5*np.sqrt(2))
+    obs = dagmc_stats.get_triangle_aspect_ratio(my_core, root_set)
+    assertAlmostEqual(exp,obs)
+    
+    exp = len(dagmc_stats.get_triangle_aspect_ratio(my_core, root_set))
```
gonuke: I think these two are backwards, at least in terms of the standard interpretation of `exp` and `obs`. 

Since you are testing `get_triangle_aspect_ratio`, its return value should be the observed result.

yqin43: I am sorry. I will also fix the backwards in test_get_area_triangle function together on Friday.

gonuke:
This PR is looking good now.  A lot of important design elements have been addressed and the structure is a great starting place and model for going forward.  Thanks for your work @yqin43

PR #55, opened 2019-11-04
Coarseness

yqin43:
## Description
Added function that calculates the coarseness of all the surfaces

## Changes
In dagmc_stats.py: added function for coarseness calculation(get_coarseness)
In generate_stats.py: added code to call the get_coarseness function

### DIFF
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -190,7 +190,7 @@ def get_tris(my_core, meshset, geom_dim):
 
 def get_tri_side_length(my_core, tri):
     """
-    get side lengths of triangle
+    Get side lengths of triangle
 
     inputs
     ------
@@ -221,7 +221,7 @@ def get_tri_side_length(my_core, tri):
 
 def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
     """
-    Gets the triangle aspect ratio (according to the equation: (abc)/(8(s-a)(s-b)(s-c)), where s = .5(a+b+c).)
+    Get the triangle aspect ratio (according to the equation: (abc)/(8(s-a)(s-b)(s-c)), where s = .5(a+b+c).)
 
     inputs
     ------
@@ -249,7 +249,7 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
 
 def get_area_triangle(my_core, meshset, geom_dim):	
     """
-    Gets the triangle area (according to the Heron's formula: 
+    Get the triangle area (according to the Heron's formula: 
     sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2)
 
     inputs
@@ -278,7 +278,8 @@ def get_area_triangle(my_core, meshset, geom_dim):
 
 def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
     """
-    Gets the coarseness of area
+    Get the coarseness of area. Coarseness is calculated by dividing 
+    surface area of a surface by number of triangles in that surface.
 
     inputs
     ------
@@ -290,8 +291,6 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
     outputs
     -------
     coarseness : (list) the coarseness for surfaces in the meshset.
-                 Coarseness is calculated by dividing surface area of
-                 a surface by number of triangles in that surface.
     """
 
     coarseness = []
--- a/dagmc_stats.py
+++ b/dagmc_stats.py
@@ -163,12 +163,12 @@ def get_tris(my_core, meshset, geom_dim):
     inputs
     ------
     my_core : a MOAB core instance
-    entity_ranges : a dictionary of the entityset ranges of each tag in a file
+    meshset : the root meshset for the file
     geom_dim : a MOAB Tag that holds the dimension of an entity.
 
     outputs
     -------
-    tris : (list)triangle entities
+    tris : a list of triangle entities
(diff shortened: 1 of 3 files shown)


test_dagmc_stats.py
```diff
@@ -99,3 +90,33 @@ def test_get_surfaces_per_volume():
...
+def test_get_triangle_aspect_ratio():
+    """
+    Tests part of the get_triangle_aspect_ratio function
+    """
+    test_input2 = "single-cube.h5m"
+    my_core = core.Core()
+    my_core.load_file(test_input2)
+    root_set = my_core.get_root_set()
+
+    assert(abs(dagmc_stats.get_triangle_aspect_ratio(my_core, root_set) - 1000*np.sqrt(2)/400/(10-5*np.sqrt(2))) < 0.01)
```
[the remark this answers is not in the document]
bradlaufenberg: I would have to agree, maybe do the calculations in a line above this and assign a variable name then call that variable in the `assert` line

dagmc_stats.py
```diff
@@ -201,3 +201,90 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
         t_a_r.append(top/bottom)
 
     return t_a_r
+
+def get_area_triangle(my_core, meshset):	
+    """
+    Gets the triangle area (according to the equation: sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2)
```
bradlaufenberg: Maybe add a short note that this is Heron's formula to give the user more detail if they are unfamiliar with it and where it comes from.

gonuke:
I think this needs to be rebased against the recently merged PR.

gonuke:
> I think this needs to be rebased against the recently merged PR.

Looks like you just fixed this - ignore...

dagmc_stats.py
```diff
@@ -270,3 +270,50 @@ def get_area_triangle(my_core, meshset, geom_dim):
...
+    meshset : a meshset containing a certain part of the mesh
+
+    outputs
+    -------
+    coarseness : (list) the coarseness for all surfaces in the meshset
+    """
+    t_p_s = {}
+    coarseness = {}
+    c_data = []
+    for surface in entity_ranges['Surfaces']:
```
gonuke: Maybe we should just pass in `entity_ranges['Surfaces']` instead of the whole `entity_ranges`?

dagmc_stats.py
```diff
@@ -270,3 +270,50 @@ def get_area_triangle(my_core, meshset, geom_dim):
...
+        t_p_s[surface] = my_core.get_entities_by_type(
+                                 surface, types.MBTRI)
+        surf_area = 0
+        for triangle in t_p_s[surface]:
+            side_lengths = []
+            s = 0
+            coord_list = []
+
+            verts = list(my_core.get_adjacencies(triangle, 0))
```
gonuke: Should you be using `get_side_lengths()` now here?

dagmc_stats.py
```diff
@@ -270,3 +270,50 @@ def get_area_triangle(my_core, meshset, geom_dim):
...
+                coords = my_core.get_coords(vert)
+                coord_list.append(coords)
+
+            for side in range(3):
+                side_lengths.append(np.linalg.norm(coord_list[side] - coord_list[side - 2]))
+                # The indices of coord_list includes the "-2" because this way each side will be matched up with both
+                # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
+
+            # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
+            s = sum(side_lengths)/2
```
gonuke: If you had a function to calculate a single triangle's area from a set of sidelengths, you could call it here (and in your other method).

dagmc_stats.py
```diff
@@ -270,3 +270,35 @@ def get_area_triangle(my_core, meshset, geom_dim):
...
+    for surface in entity_ranges:
+        t_p_s[surface] = my_core.get_entities_by_type(
+                                 surface, types.MBTRI)
+        surf_area = 0
+        for tri in t_p_s[surface]:	
+            side_lengths = get_tri_side_length(my_core, tri)
+            # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
+            s = sum(side_lengths)/2
+            s = np.sqrt(s * np.prod(s - side_lengths))
+            surf_area += s
```
kkiesling: Rather than recalculate this all again, isn't this just the triangle area calculation that is done already in a different function? Can you call that function to do this work? You will need to pass in a single surface to the triangle area function and get the list of triangle areas for that surface. Something like this might work better:

```
for surf in geometry:
    triangle_areas = get_triangle_area(surf)  # list would return a list of the triangle areas for the surface of interest
    calculated_coarseness = len(triangle_areas) / sum(triangle_areas) # this is equivalent to number of triangles on the  surface / the total area of the surface.

You do the right coarseness calculation in line 302, but I think you could use some functions you already made in your last PR to calculate the area.

kkiesling:
@yqin43 can you rebase against master to get your most recent PRs into this one? That may require you to resolve a conflict in the test file. Let me know if you need help resolving the conflict.

kkiesling:
I think comments from @gonuke have been addressed, but I'll let him approve before I merge.

gonuke:
LGTM - thanks for your work on this @yqin43

PR #59, opened 2020-02-28
Roughness

yqin43:
## Description
5 functions  calculating roughness are added:
(roughness is calculated based on **part 3.2 Local Roughness** of  article "**A fast roughness-based approach to the assessment of 3D mesh visual quality**")

get_beta_angles: get the beta angles (See **fig 3**)
get_angles: get angles of a triangle(the first element in the list "angles" is always the alpha angle)(See **fig 3**)
gaussian_curvature: calculate gaussian curvature value of given vertex (See **formula 1**)
get_local_roughness: calculate local roughness value of given vertex (See **formula 2**)
get_roughness: return local roughness values for all the vertices (See **formula 3**)

## Other Information
Reference: **A fast roughness-based approach to the assessment of 3D mesh visual quality**
https://www.sciencedirect.com/science/article/pii/S0097849312001203


dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+    pair_tris : triangle pair
+    common_vert : common vertices in the triangle pair
+    beta_angles: list that stores the beta angles of given triangle pair
+    """
+    vert1 = common_vert[0]
+    vert2 = common_vert[1]
+    side_lengths = []
+    
+    for m in range(2):
```
gonuke: Why not:
```suggestion
    for tri in pair_tris:
```
and corresponding changes below....

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        adj_verts = my_core.get_adjacencies(pair_tris[m], 0, op_type=1)
+        for n in range(3):
```
```suggestion
        for vert in adj_verts:
```
and corresponding changes below

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+def get_beta_angles(my_core, pair_tris, common_vert, beta_angles):
+    """
+    Gets the beta angles of given triangle pairs
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    pair_tris : triangle pair
+    common_vert : common vertices in the triangle pair
```
```suggestion
    common_vert : entity handles of common vertices in the triangle pair
```

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+            if adj_verts[n] != vert1 and adj_verts[n] != vert2:
+                other_vert = adj_verts[n]        
+        side_lengths.append(np.linalg.norm(my_core.get_coords(vert1)-my_core.get_coords(vert2)))
+        side_lengths.append(np.linalg.norm(my_core.get_coords(vert1)-my_core.get_coords(other_vert)))
+        side_lengths.append(np.linalg.norm(my_core.get_coords(vert2)-my_core.get_coords(other_vert)))
+        beta_angles.append(np.arccos((side_lengths[1] * side_lengths[1]
+                            + side_lengths[2] * side_lengths[2]
+                            - side_lengths[0] * side_lengths[0])
+                            /(2.0 * side_lengths[1] * side_lengths[2])))
+
```
gonuke: I wonder if there is a formulation of this that doesn't require you to identify the `other_vert` and reuse your `get_tri_side_lengths` method
```
shared_side_length = np.linalg.norm(my_core.get_coords(vert1) - my_core.get_coords(vert2))
side_lengths = get_tri_side_lengths(my_core, tri)
cos_beta_angle = (np.dot(side_lengths,side_lengths) - 2*shared_side_length*shared_side_length)/
                            (2.0 * np.prod(side_lengths)/shared_side_length)
beta_angles.append(np.arccos(cos_beta_angle))
```

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    angles = []
+    side_lengths = []
+    coord_list = []
+    index = 3
+    if vert != None:
+        verts = my_core.get_adjacencies(tri, 0, op_type=1)
+        for i in range(len(verts)):
+            if verts[i] == vert:
+                index = i
+        for vert in verts:
```
gonuke: careful reusing the variable `vert` that is in your parameter list....

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+def get_angles(my_core, tri, vert = None):
+    """
+    Gets the angles of the given triangle
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    tri: triangle
+    vert: the vertex which generates the triangle by get_adj
```
```suggestion
    vert: the entity handle of the vertex which generates the triangle by get_adj
```

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    side_lengths = []
+    coord_list = []
+    index = 3
+    if vert != None:
+        verts = my_core.get_adjacencies(tri, 0, op_type=1)
+        for i in range(len(verts)):
+            if verts[i] == vert:
+                index = i
+        for vert in verts:
+            coords = my_core.get_coords(vert)
```
gonuke: Is it possible to call `get_coords` with a list of vertices and get back all the coordinates at once?

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        side_lengths.append(np.linalg.norm(coord_list[index]-coord_list[index-1]))
+        side_lengths.append(np.linalg.norm(coord_list[index]-coord_list[index-2]))
+        side_lengths.append(np.linalg.norm(coord_list[index-1]-coord_list[index-2]))
+    else:
+        side_lengths = get_tri_side_length(my_core, tri)
+    for side in range(3):
+        angles.append(np.arccos((side_lengths[side] * side_lengths[side]
+                            + side_lengths[side - 2] * side_lengths[side - 2]
+                            - side_lengths[side - 1] * side_lengths[side - 1])
+                            /(2.0 * side_lengths[side] * side_lengths[side - 2])))
```
gonuke: Almost certainly better here to do something like:
```
side_lengths_sq = np.dot(side_lengths,side_lengths)
side_length_prod = np.prod(side_lengths)
for this_side_length in side_lengths:
    cos_angle = (side_lengths_sq - 2.0*this_side_length*this_side_length)/(2.0 * side_length_prod/this_side_length)
    angles.append(np.arccos(cos_angle))
```

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+    vert : vertex
+
+    outputs
+    -------
+    gc : the gaussian curvature of the vert
+    """ 
+    
+    tris = my_core.get_adjacencies(vert, 2)
+    alpha_angles = 0
```
gonuke: Should this be initialized as a list since `get_angles` returns a list?

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+
+def gaussian_curvature(my_core, vert):
+    """
+    Gets gaussian curvature of a vert
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    vert : vertex
```
```suggestion
    vert : entity handle of vertex being assessed
```

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+    adj_tris = my_core.get_adjacencies(vert, 2, op_type=0)
+    
+    #pair tris
+    pair_tris = []
+    tri_vert = []
+    common_vert = []
+    for tri in adj_tris:
+        tri_vert.append(my_core.get_adjacencies(tri, 0, op_type=1))
+    for tri1 in tri_vert:
```
gonuke: I have a hunch that using sets may make a lot of this simpler...

dagmc_stats.py
gonuke: furthermore, you might not need to define `vert1` and `vert2`, but just use the `common_vert[0]` and `common_vert[1]` since these are now the only place they are used.

dagmc_stats.py
gonuke: Perhaps we should discuss or study the pyMOAB methods that are designed to do some of this.  Some of what you are doing here is capability built in to MOAB.

gonuke:
Thanks for this complicated new addition!  It looks like a lot of work and very valuable to @kkiesling's progress.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+    for i in range(len(adj_verts)):
+        sum_d_gc += d[i] * gc_j[i]
+    lr = np.abs(gc_i - sum_d_gc / sum(d))
+    return lr
+
+
+def get_roughness(my_core, native_ranges):
+    """
+    Gets the roughness of area
```
kkiesling: Add a reference to the paper that this method is based on.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    native_ranges : a dictionary containing ranges for each native type in the file (VERTEX, TRIANGLE, ENTITYSET)
```
kkiesling: PEP8: line exceeds maximum line length for PEP8 style guide

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+    outputs
+    -------
+    roughness: (list) the roughness for all surfaces in the meshset.
```
kkiesling: Do you return roughness for each surface or each vertex? It looks like you do each vertex.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    Gets local roughness of a vert
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    vert : vertex
+
+    outputs
+    -------
+    lr : the laplacian value of the vert
```
kkiesling:
```suggestion
    lr : the local roughness value of the vert
```

dagmc_stats.py
kkiesling: I would be interested in talking more about how to streamline this because we had a hard time coming up with a method to identify which entities were the relevant ones for calculating the beta angles.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                pair = []
+                pair.append(tri1)
+                pair.append(tri2)
+                exist = False
+                for p in pair_tris:
+                    if (p[0] == tri1 and p[1] == tri2) or (p[1] == tri1 and p[0] == tri2):
+                        exist = True
+                if not exist:
+                    common_vert.append(same_vert)
+                    pair_tris.append(pair)
```
kkiesling: Can you add more comments to this section so that it is easier to follow? Such as why each for loop exists and what the intention is, what you are checking with each if-statement, and what the end goal is of this whole process.

dagmc_stats.py
kkiesling: This would be a good SW design meeting conversation (cc @bam241)

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+def get_beta_angles(my_core, pair_tris, common_vert, beta_angles):
+    """
+    Gets the beta angles of given triangle pairs
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    pair_tris : triangle pair
+    common_vert : common vertices in the triangle pair
+    beta_angles: list that stores the beta angles of given triangle pair
```
kkiesling: This function doesn't ever return anything. Are you sure you are properly getting the beta angles to use after calling this function?

dagmc_stats.py
kkiesling: I'd like to have a larger software design conversation in person with more people involved on this calculation of the beta angles.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                exist = False
+                for p in pair_tris:
+                    if (p[0] == tri1 and p[1] == tri2) or (p[1] == tri1 and p[0] == tri2):
+                        exist = True
+                if not exist:
+                    common_vert.append(same_vert)
+                    pair_tris.append(pair)
+    #d
+    for p in range(len(pair_tris)):
+        get_beta_angles(my_core, pair_tris[p], common_vert[p],beta_angles)
```
kkiesling: Double check that you are actually getting values returned in this function. My suspicion is that beta_angles is still an empty list after this.

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        angles.append(np.arccos((side_lengths[side] * side_lengths[side]
+                            + side_lengths[side - 2] * side_lengths[side - 2]
+                            - side_lengths[side - 1] * side_lengths[side - 1])
+                            /(2.0 * side_lengths[side] * side_lengths[side - 2])))
+    return angles
+
+
+def gaussian_curvature(my_core, vert):
+    """
+    Gets gaussian curvature of a vert
```
kkiesling: reference the paper and the equation used to calculate GC

dagmc_stats.py
```diff
@@ -300,3 +300,184 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+    pair_tris : triangle pair
+    common_vert : entity handles of common vertices in the triangle pair
+    beta_angles: list that stores the beta angles of given triangle pair
+    """
+    vert1 = common_vert[0]
+    vert2 = common_vert[1]
+    side_lengths = []
+    
+    for tri in pair_tris:
```
gonuke: Having made this recommended change, you need to consider how they propagate below.  For example `pair_tris[m]` is no longer defined because you don't have an index variable `m` anymore.  You may want to review that carefully.

yqin43: Yes. I changed them correspondingly. Thank you!

dagmc_stats.py
yqin43: In order to know which angle is the alpha or beta angle, I need to preserve the order of the angles. This seems to be calculating the cos values in other order and therefore does not work.

dagmc_stats.py
yqin43: I am calculating the sum of all the alpha angles so 0 works. I will change the name of this variable into sum_alpha_angles. Thank you!

gonuke: but `get_angles()` returns a list.  I am not sure that it will behave how you imagine?

```
alpha_angles += get_angles(...)
```
will look something like
```
0 += [angle1, angle2, angle3, ...]
```

yqin43: I am only getting the first element of the list returned by get_angles()(the alpha angle). Then I compute the sum by sum_alpha_angles += get_angles[0] so it works. sum_alpha_angles only stores value of the sum of  all the alpha angles.

yqin43:
@gonuke @kkiesling Thank you for your suggestions and I have edited them except the ones corresponding to how to get the beta angles. I have also updated the test file.

I would like to talk about the beta angle problem during the software meeting as Kalin suggested. Currently it seems to be pretty complicated.

test_dagmc_stats.py
```diff
@@ -127,3 +128,22 @@ def test_get_area_triangle():
...
+    test_input2 = "single-cube.h5m"
+    my_core = core.Core()
+    my_core.load_file(test_input2)
+    root_set = my_core.get_root_set()
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+    native_ranges = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)
+    
+    exp = 0
+    obs = get_roughness(my_core, native_ranges)[0]
+    assertEqual(True, np.abs(obs-exp) < 0.01)
```
kkiesling: you can use `assertAlmostEqual()` here instead

test_dagmc_stats.py
```diff
@@ -127,3 +128,22 @@ def test_get_area_triangle():
...
+    """
+    test_input2 = "single-cube.h5m"
+    my_core = core.Core()
+    my_core.load_file(test_input2)
+    root_set = my_core.get_root_set()
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+    native_ranges = dagmc_stats.get_native_ranges(my_core, root_set, entity_types)
+    
+    exp = 0
+    obs = get_roughness(my_core, native_ranges)[0]
```
kkiesling: this function returns an entire list, correct? We should be testing if the entire list matches our expected result.

test_dagmc_stats.py
```diff
@@ -127,3 +128,22 @@ def test_get_area_triangle():
...
+    assertEqual(True, np.abs(obs-exp) < 0.01)
+
+    exp = my_core.get_entities_by_type(root_set, types.MBVERTEX).size()
+    obs = len(get_roughness(my_core, native_ranges))
+    assertEqual(exp,obs)
```
kkiesling: this assert of the lengths seems unnecessary if we can test that the lists are equal above.

dagmc_stats.py
```diff
@@ -300,3 +300,178 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                                                -my_core.get_coords(verts[1])))
+    side_lengths.append(np.linalg.norm(my_core.get_coords(verts[0])
+                                                -my_core.get_coords(vert)))
+    side_lengths.append(np.linalg.norm(my_core.get_coords(verts[1])
+                                                -my_core.get_coords(vert)))
+    beta_angle1 = np.arccos((side_lengths[0] * side_lengths[0]
+                            + side_lengths[2] * side_lengths[2]
+                            - side_lengths[1] * side_lengths[1])
+                            /(2.0 * side_lengths[0] * side_lengths[2]))                                            
+    vert_dic.get(str(lst1)).append(beta_angle1)
```
kkiesling:
```suggestion
    vert_dic[lst1].append(beta_angle1)
```

dagmc_stats.py
```diff
@@ -300,3 +300,178 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+    vert : entity handle of the central vertex
+    verts : entity handles of the adjacent vertices of vert
+    vert_dic : a dictionary storing common vertex:beta angles
+    """
+    
+    verts.remove(vert)
+    side_lengths = []
+    lst1 = sorted([vert, verts[0]])
+    lst2 = sorted([vert, verts[1]])
```
kkiesling:
```suggestion
    lst1 = str(sorted([vert, verts[0]]))
    lst2 = str(sorted([vert, verts[1]]))
```
You only use these variables as a dictionary key, so they can be converted to strings right away. See comments below for corresponding code changes.

dagmc_stats.py
```diff
@@ -300,3 +300,178 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                            + side_lengths[2] * side_lengths[2]
+                            - side_lengths[1] * side_lengths[1])
+                            /(2.0 * side_lengths[0] * side_lengths[2]))                                            
+    vert_dic.get(str(lst1)).append(beta_angle1)
+    
+    beta_angle2 = np.arccos((side_lengths[0] * side_lengths[0]
+                            + side_lengths[1] * side_lengths[1]
+                            - side_lengths[2] * side_lengths[2])
+                            /(2.0 * side_lengths[0] * side_lengths[1]))
+    vert_dic.get(str(lst2)).append(beta_angle2)
```
kkiesling:
```suggestion
    vert_dic[lst2].append(beta_angle2)
```

dagmc_stats.py
```diff
@@ -300,3 +300,178 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    side_lengths = []
+    lst1 = sorted([vert, verts[0]])
+    lst2 = sorted([vert, verts[1]])
+    
+    side_lengths.append(np.linalg.norm(my_core.get_coords(verts[0])
+                                                -my_core.get_coords(verts[1])))
+    side_lengths.append(np.linalg.norm(my_core.get_coords(verts[0])
+                                                -my_core.get_coords(vert)))
+    side_lengths.append(np.linalg.norm(my_core.get_coords(verts[1])
+                                                -my_core.get_coords(vert)))
```
kkiesling: readability suggestion: add a blank line after line 327

dagmc_stats.py
```diff
@@ -300,3 +300,178 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+
+def get_angles(my_core, tri, vert = None):
+    """
+    Gets the angles of the given triangle
+
+    inputs
+    ------
+    my_core : a MOAB Core instance
+    tri: triangle
```
kkiesling: specify that this is a triangle entity handle

dagmc_stats/dagmc_stats.py
```diff
@@ -199,7 +199,7 @@ def get_tri_side_length(my_core, tri):
     side_lengths : a list of side lengths of triangle
```
kkiesling, line 199: Docstring needs updating because `side_lengths` is now a dictionary. This should describe the dict structure.

dagmc_stats/dagmc_stats.py
```diff
@@ -210,7 +210,7 @@ def get_tri_side_length(my_core, tri):
         coord_list.append(coords)
 
     for side in range(3):
-        side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
+        side_lengths.update({verts[side-1]:np.linalg.norm(coord_list[side]-coord_list[side-2])})
         # The indices of coord_list includes the "-2" because this way each side will be matched up with both
         # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
```
kkiesling, line 215: PEP8 style guide: these lines should only be a max of 79 characters.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
         coarseness.append(len(surf_area)/sum(surf_area))
 
     return coarseness
+
+
+def get_tri_vert_data(my_core, native_ranges):
+    """
+    Build a numpy strcutured array to store triangle and vertex related data in the form of
```
kkiesling, line 304: PEP8: The docstring should start on the same line as the `"""`. And line length of comments and doctrings should be no longer than 72 characters.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    outputs
+    -------
+    tri_vert_data : a numpy structured array that stores the triangle and vertex related data
+    all_verts : (list) all the vertices that are connected to triangle in the geometry
+    """
+    
+    all_tris = native_ranges[types.MBTRI]
+    all_verts = []
+    
+    tri_vert_struct = np.dtype({'names':['tri','vert','angle','side_length'], 'formats':[np.uint64, np.uint64, np.float64, np.float64]})
```
kkiesling, line 321: PEP8: line length should be no more than 79 characters. And a space is needed after the commas `,` and colons `:`.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+    inputs
+    ------
+    vert_i : vertex entity handle
+    tri_vert_data : numpy structured array that stores the triangle and vertex related data
+    
+    outputs
+    -------
+    gc : gaussian curvature value of the vertex
+
```
kkiesling, line 375: PEP8: remove blank line

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    ------
+    vert_i : vertex entity handle
+    tri_vert_data : numpy structured array that stores the triangle and vertex related data
+    
+    outputs
+    -------
+    gc : gaussian curvature value of the vertex
+    
+    """
+
```
kkiesling, line 377: PEP8: remove blank line

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    for tri in all_tris:
+        side_lengths = get_tri_side_length(my_core, tri)   # {vert : side_length}
+        side_length_sum_sq = sum(map(lambda i : i * i, side_lengths.values()))
+        side_length_prod = np.prod(side_lengths.values())
+        verts = list(my_core.get_adjacencies(tri, 0, op_type=1))
+        for vert_i in verts:
+            if vert_i not in all_verts:
+                all_verts.append(vert_i)
+            side_i = side_lengths[vert_i]
+            d_i = np.arccos((side_length_sum_sq - 2 * (side_i**2)) * side_i/(2 * side_length_prod))
```
kkiesling, line 333: PEP8: line length

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    """
+    
+    all_tris = native_ranges[types.MBTRI]
+    all_verts = []
+    
+    tri_vert_struct = np.dtype({'names':['tri','vert','angle','side_length'], 'formats':[np.uint64, np.uint64, np.float64, np.float64]})
+    tri_vert_data = np.zeros(0,dtype=tri_vert_struct)
+    
+    for tri in all_tris:
+        side_lengths = get_tri_side_length(my_core, tri)   # {vert : side_length}
```
kkiesling, line 325: PEP8: line length

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        side_length_sum_sq = sum(map(lambda i : i * i, side_lengths.values()))
+        side_length_prod = np.prod(side_lengths.values())
+        verts = list(my_core.get_adjacencies(tri, 0, op_type=1))
+        for vert_i in verts:
+            if vert_i not in all_verts:
+                all_verts.append(vert_i)
+            side_i = side_lengths[vert_i]
+            d_i = np.arccos((side_length_sum_sq - 2 * (side_i**2)) * side_i/(2 * side_length_prod))
+            bar = np.array((tri,vert_i,d_i,side_i),dtype=[('tri',np.uint64), ('vert',np.uint64), ('angle',np.float64), ('side_length',np.float64)])
```
kkiesling, line 334: PEP8: line length

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+            tri_vert_data = np.append(tri_vert_data,bar)
```
kkiesling, line 335:
```suggestion
            tri_vert_data = np.append(tri_vert_data, bar)
```
PEP8 space change (there are other locations as well)

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+    all_tris = native_ranges[types.MBTRI]
+    all_verts = []
+    
+    tri_vert_struct = np.dtype({'names':['tri','vert','angle','side_length'], 'formats':[np.uint64, np.uint64, np.float64, np.float64]})
+    tri_vert_data = np.zeros(0,dtype=tri_vert_struct)
+    
+    for tri in all_tris:
+        side_lengths = get_tri_side_length(my_core, tri)   # {vert : side_length}
+        side_length_sum_sq = sum(map(lambda i : i * i, side_lengths.values()))
```
kkiesling, line 326:
```suggestion
        side_length_sum_sq = sum(map(lambda i: i**2, side_lengths.values()))
```

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+            side_i = side_lengths[vert_i]
+            d_i = np.arccos((side_length_sum_sq - 2 * (side_i**2)) * side_i/(2 * side_length_prod))
+            bar = np.array((tri,vert_i,d_i,side_i),dtype=[('tri',np.uint64), ('vert',np.uint64), ('angle',np.float64), ('side_length',np.float64)])
+            tri_vert_data = np.append(tri_vert_data,bar)
+    return tri_vert_data, all_verts
+
+
+def get_gaussian_curvature(my_core, native_ranges, all_verts, tri_vert_data):
+    """
+    Get gaussian curvature values of all non-isolated vertices
```
kkiesling, line 341: See other comments about docstring PEP8 formatting

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+            if vert_i not in all_verts:
+                all_verts.append(vert_i)
+            side_i = side_lengths[vert_i]
+            d_i = np.arccos((side_length_sum_sq - 2 * (side_i**2)) * side_i/(2 * side_length_prod))
+            bar = np.array((tri,vert_i,d_i,side_i),dtype=[('tri',np.uint64), ('vert',np.uint64), ('angle',np.float64), ('side_length',np.float64)])
+            tri_vert_data = np.append(tri_vert_data,bar)
+    return tri_vert_data, all_verts
+
+
+def get_gaussian_curvature(my_core, native_ranges, all_verts, tri_vert_data):
```
kkiesling, line 339: check line lengths on lines of code in this function

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+    native_ranges : a dictionary containing ranges for each native type in the file (VERTEX, TRIANGLE, ENTITYSET)
+    all_verts : all the vertices that are connected to triangle in the geometry
+    tri_vert_data : numpy structured array that stores the triangle and vertex related data
+    
+    outputs
+    -------
+    gc_all : dictionary in the form of vertex : gaussian curvature value of the vertex
+    """
+
```
kkiesling, line 354: remove blank line

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    ------
+    vert_i : vertex entity handle
+    gc_all : dictionary in the form of vertex : gaussian curvature value of the vertex
+    tri_vert_data : numpy structured array that stores the triangle and vertex related data
+    my_core : a MOAB Core instance
+    
+    outputs
+    -------
+    Lri : local roughness value of the vertex
+
```
kkiesling, line 400: remove

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    gc_all : dictionary in the form of vertex : gaussian curvature value of the vertex
+    tri_vert_data : numpy structured array that stores the triangle and vertex related data
+    my_core : a MOAB Core instance
+    
+    outputs
+    -------
+    Lri : local roughness value of the vertex
+    
+    """
+
```
kkiesling, line 402: remove

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    
+    outputs
+    -------
+    Lri : local roughness value of the vertex
+    
+    """
+    
+    DIJgc_sum = 0
+    Dii_sum = 0
+    vert_j_list = list(my_core.get_adjacencies(my_core.get_adjacencies(vert_i,2,op_type=0),0,op_type=1))
```
kkiesling, line 405:
```suggestion
    vert_j_list = list(my_core.get_adjacencies(my_core.get_adjacencies(vert_i, 2, op_type=0), 0, op_type=1))
```

kkiesling: the line is still too long with this suggestion FYI

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        tri_j_list = my_core.get_adjacencies(vert_j, 2, op_type=0)
+        tri_ij_list = list(set(tri_i_list) & set(tri_j_list))
+        # get vertices that are in triangles in tri_ij_list
+        # there will be six elements representing four vertices
+        # vert_i and vert_j will appear twice
+        four_vert_0 = tri_vert_data[tri_vert_data['tri'] == tri_ij_list[0]]
+        four_vert_1 = tri_vert_data[tri_vert_data['tri'] == tri_ij_list[1]]
+        four_vert = np.concatenate([four_vert_0, four_vert_1])
+        
+        #get beta angles
```
kkiesling, line 419:
```suggestion
        # get beta angles
```

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+        beta_angles = []
+        for entry in four_vert:
+            matched = four_vert[four_vert['vert'] == entry['vert']]
+            if len(matched) < 2:
+                beta_angles.append(entry['angle'])
```
kkiesling, line 424: I know we discussed this logic in person, but can you also add some comments describing what is happening here so others understand.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +297,157 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                
+        Dij = 0.5*(1/np.tan(beta_angles[0]) + 1/np.tan(beta_angles[1]))
+        DIJgc_sum += ( Dij * gc_all[vert_j] )
```
kkiesling, line 427:
```suggestion
        DIJgc_sum += (Dij * gc_all[vert_j])
```

dagmc_stats/generate_stats.py
```diff
@@ -190,6 +198,11 @@ def collect_statistics(my_core, root_set, tar_meshset, display_options):
         data[c_key] = dagmc_stats.get_coarseness(my_core, root_set, 
                                                                 entityset_ranges['Surfaces'], dagmc_tags['geom_dim'])
         stats[c_key] = get_stats(data[c_key])
+    
+    if display_options['R']:
+        r_key = 'R'
```
kkiesling, line 203: is `r_key` used anywhere else? I think you can just use `data['R']` and `stats['R']`.

yqin43: Since the previous options are added in this way, I added like this to keep it consistent

yqin43:
Code in dagmc_stats.py has been edited according to PEP8 style and tests for new functions are added. Please check these changes. Thanks!

tests/test_dagmc_stats.py
```diff
@@ -104,39 +108,71 @@ def test_get_triangle_aspect_ratio(self):
...
+        obs = len(tri_vert_data)
+        self.assertEqual(exp,obs)
+        
+        exp = 8
+        obs = len(all_verts)
+        self.assertEqual(exp,obs)
+        
+        for i in range(8):
+            obs = len(tri_vert_data[tri_vert_data['vert'] == all_verts[i]])
+            self.assertTrue(obs == 4 or obs == 5)
```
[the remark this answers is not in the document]
yqin43: I tried to print out tri_vert_data's vert values and the value of obs will be 5 5 4 4 5 4 5 4 in order. Should I edit the test to the form where I test the first, second, fifth and seventh values are 5s and the rest are 4s, or should I just leave it as it is now?

tests/test_dagmc_stats.py
```diff
@@ -104,39 +108,84 @@ def test_get_triangle_aspect_ratio(self):
...
+        obs = len(tri_vert_data)
+        self.assertEqual(exp, obs)
+        
+        exp = 8
+        obs = len(all_verts)
+        self.assertEqual(exp, obs)
+        
+        for i in range(8):
+            obs = len(tri_vert_data[tri_vert_data['vert'] == all_verts[i]])
+            self.assertTrue(obs == 4 or obs == 5)
```
kkiesling, line 148:
```suggestion
        exp = [5, 5, 4, 4, 5, 4, 5, 4]
        for i in range(8):
            obs = len(tri_vert_data[tri_vert_data['vert'] == all_verts[i]])
            self.assertTrue(obs == exp[i])
```

kkiesling:
@yqin43 - can you change this PR so that you are requesting merging into the `main` branch rather than the `master`? You should be able to click the clipboard button at the top of the page to edit the branch.

kkiesling:
never mind- I have permission to change the base branch so I updated it.

tests/test_dagmc_stats.py
```diff
@@ -104,39 +108,71 @@ def test_get_triangle_aspect_ratio(self):
...
+        gc_all = ds.get_gaussian_curvature(my_core_2, native_ranges_2,all_verts, tri_vert_data)
+        
+        exp = 8
+        obs = len(gc_all)
+        self.assertEqual(exp,obs)
+        
+        exp = 0.5*np.pi
+        for i in range(8):
+            obs = gc_all[all_verts[i]]
+            self.assertAlmostEqual(exp,obs)
```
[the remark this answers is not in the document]
kkiesling: @yqin43 can you accept these few pep8 changes?

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+        # get beta angles
+        # remove the verts that are shared by two triangles (that will appear
+        # twice)
+        beta_angles = []
+        for entry in four_vert:
+            matched = four_vert[four_vert['vert'] == entry['vert']]
+            if len(matched) < 2:
+                beta_angles.append(entry['angle'])
+
```
gonuke, line 460: To make the best use out of the structured array try this:
```suggestion
        select_tris = (tri_vert_data['tri'] == tri_ij_list[0]) | (tri_vert_data['tri'] == tri_ij_list[1])
        exclude_verts = (tri_vert_data['vert'] != vert_i) & (tri_vert_data['vert'] != vert_j)
        beta_angles = tri_vert_data[ select_tris & exclude_verts ]['angle']
```
The first line creates a "mask" of all the rows in which the 'tri' is `tri_ij_list[0]` or `tri_ij_list[1]`.

The second line creates a "mask" of all the rows in which the 'vert' is NOT `vert_i` and NOT `vert_j`.

The third line combines those masks to get the two rows you want, and only takes the 'angle' column.

I tested this syntax in an interpreter window with some made up data, and I'm pretty sure this is the right syntax for your case.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    in the file (VERTEX, TRIANGLE, ENTITYSET)
+
+    outputs
+    -------
+    roughness : (list) the roughness for all surfaces in the meshset
+    """
+    tri_vert_data, all_verts = get_tri_vert_data(my_core, native_ranges)
+    gc_all = get_gaussian_curvature(my_core, native_ranges, all_verts,
+                                    tri_vert_data)
+    roughness = []
```
gonuke, line 484: Should this be a numpy array (should we always prefer numpy arrays)?  Perhaps review other places in the code where lists are used and decide if they should numpy arrays?

gonuke: Related question - should we pre-allocate this to be long enough for all data and then populate it rather than appending.  This can lead to memory resize events that can be slow when the size of `all_verts` gets large (e.g. O(10^6))

tests/test_dagmc_stats.py
```diff
@@ -8,13 +8,19 @@
...
 test_input = "tests/3vols.h5m"
-
 my_core = core.Core()
 my_core.load_file(test_input)
-
 root_set = my_core.get_root_set()
+
 entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
+test_input_2 = "tests/single-cube.h5m"
```
gonuke, line 17: I always get a little uneasy when I see variable names with numbers in them to distinguish between cases.

A handy solution in python would be to have something like:
```
test_env = [ { 'input_file' : 'test/3vols.h5m'}, {'input_file' : 'tests/single-cube.h5m'} ]

for env in test_env:
     env['core'] = core.Core()
     env['core'].load_file(env['input_file'])
     env['root_set'] = env['core'].get_root_set()
     env['native_ranges'] = ds.get_native_ranges(env['core'], env['root_set'], entity_types)
     env['dagmc_tags'] = ds.get_dagmc_tags(env['core'])
```

Then in each test, you can select which `env` you want to use....

tests/test_dagmc_stats.py
```diff
@@ -71,20 +77,18 @@ def test_get_triangles_per_surface(self):
...
-
-        #triangles = my_core.get_entities_by_type(root_set, types.MBTRI).size()
-        #assert(sum(t_p_s_data) == triangles)
+            my_core_2, entityset_ranges)
+            
+        assert(len(t_p_s_data) == 6)
+        
+        assert(sum(t_p_s_data.values()) == 12)
+        
+        for v in t_p_s_data.values():
```
gonuke, line 90: numpy has a way to compare two arrays: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

tests/test_dagmc_stats.py
```diff
@@ -104,39 +108,85 @@ def test_get_triangle_aspect_ratio(self):
...
+        """
+        tri_vert_data, all_verts = ds.get_tri_vert_data(my_core_2, native_ranges_2)
+        gc_all = ds.get_gaussian_curvature(my_core_2, native_ranges_2,all_verts, tri_vert_data)
+        
+        exp = 8
+        obs = len(gc_all)
+        self.assertEqual(exp, obs)
+        
+        exp = 0.5*np.pi
+        for i in range(8):
```
gonuke, line 163: numpy has an array for an approx comparison of two array: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    -------
+    tri_vert_data : a numpy structured array that stores the triangle
+    and vertex related data
+    all_verts : (list) all the vertices that are connected to
+    triangle in the geometry
+    """
+    all_tris = native_ranges[types.MBTRI]
+    all_verts = []
+
+    tri_vert_struct = np.dtype({'names': ['tri', 'vert', 'angle',
```
gonuke, line 343: Maybe this `dtype` should be defined at global scope in case you want to use it elsewhere.  You don't currently use it elsewhere, but the structured array is used elsewhere.  I can imagine opportunities to invoke this `dtype` in the future in other methods.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+    outputs
+    -------
+    tri_vert_data : a numpy structured array that stores the triangle
+    and vertex related data
+    all_verts : (list) all the vertices that are connected to
+    triangle in the geometry
+    """
+    all_tris = native_ranges[types.MBTRI]
+    all_verts = []
```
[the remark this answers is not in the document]
kkiesling: We have just tried doing this and it seems to have messed up a lot more indexing and ordering.. Not really sure why that is, but I think for now it is better to leave it as is.

dagmc_stats/dagmc_stats.py
```diff
@@ -210,15 +227,20 @@ def get_tri_side_length(my_core, tri):
...
-        side_lengths.append(np.linalg.norm(coord_list[side]-coord_list[side-2]))
-        # The indices of coord_list includes the "-2" because this way each side will be matched up with both
-        # other sides of the triangle (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
+        side_lengths.update({verts[side-1]:
+                             np.linalg.norm(coord_list[side] -
+                             coord_list[side-2])})
+        # The indiex of vert corresponds with the vert opposite to the side
+        # The indices of coord_list includes the "-2" because this way each
+        # side will be matched up with both other sides of the triangle
+        # (IDs: (Side 0, Side 1), (Side 1, Side 2), (Side 2, Side 0))
```
gonuke, line 236: Since I'm the one most anxious about understanding this in the future, I'll make a specific recommendation on this comment.
```suggestion
        # Although it may not be intuitive, the indexing of these lists takes 
        # advantage of python's indexing syntax to rotate through 
        # the `verts` of the triangle while simultaneously referencing the side
        # opposite each of the `verts` by the coordinates of the vertices that
        # define that side:
        #    side       side-1   index(side-1)     side-2   index(side-2)
        #     0               -1           2                      -2             1
        #     1                0           0                      -1             2
        #     2                1           1                       0             0
```

dagmc_stats/dagmc_stats.py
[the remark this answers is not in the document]
gonuke: hmmm  ... do you depend on `all_verts` and `tri_vert_data` having the same order?

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    my_core : a MOAB Core instance
+
+    outputs
+    -------
+    Lri : local roughness value of the vertex
+    """
+    DIJgc_sum = 0
+    Dii_sum = 0
+    vert_j_list = list(my_core.get_adjacencies(my_core.get_adjacencies(
+        vert_i, 2, op_type=0), 0, op_type=1))
```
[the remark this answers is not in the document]
kkiesling: Ya that doesn't work. You first have to get the triangles connected to vert_i and then the vertices connected to the triangles. Vertices are not directly connected to each other.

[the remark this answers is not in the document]
gonuke: Sorry for the wild goose chase.

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +316,172 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+                all_verts.append(vert_i)
+            side_i = side_lengths[vert_i]
+            d_i = np.arccos((side_length_sum_sq - 2 * (side_i**2)) * side_i /
+                            (2 * side_length_prod))
+            bar = np.array((tri, vert_i, d_i, side_i),
+                           dtype=[('tri', np.uint64),
+                                  ('vert', np.uint64),
+                                  ('angle', np.float64),
+                                  ('side_length', np.float64)])
+            tri_vert_data = np.append(tri_vert_data, bar)
```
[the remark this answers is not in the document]
kkiesling: Do you have any idea how to properly preallocate a structured array? We tried: `tri_vert_data = np.zeros(len(all_tris)*3, dtype=tri_vert_struct)`. And then we used index flattening to get the correct index. But that does not work. I don't think we know the proper way to pre-allocate when it's a structured array.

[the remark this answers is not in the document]
kkiesling: Screen Shot 2020-07-20 at 10 01 14 PM
Also tried this - doesn't seem to be allocating properly or we are not properly indexing/inserting the `bar` into the structured array.
(google is failing me on how to do this now)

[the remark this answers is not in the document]
gonuke: I think you want
`tri_vert_data = np.zeros((len(all_tris)*3,4), dtype=tri_vert_struct)`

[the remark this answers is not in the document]
gonuke: Or better?
`tri_vert_data = np.zeros((len(all_tris)*3,len(tri_vert_struct)), dtype=tri_vert_struct)`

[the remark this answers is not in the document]
gonuke: No - that's not right - this is better:
`tri_vert_data = np.zeros(len(all_tris)*3, dtype=tri_vert_struct)`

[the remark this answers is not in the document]
kkiesling: @yqin43 - so after some trial and error, the original way we allocated should be correct:
`tri_vert_data = np.zeros(len(all_tris)*3, dtype=tri_vert_struct)`

Can you run just the tests for this function and see if they pass?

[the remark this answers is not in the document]
gonuke: Regarding the last comment.... conceptually, a NUMPY structured array is a 1D array of structures, where each structure has some dtype

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +324,153 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+
+    outputs
+    -------
+    tri_vert_data : a numpy structured array that stores the triangle
+    and vertex related data
+    all_verts : (list) all the vertices that are connected to
+    triangle in the geometry
+    """
+    all_verts =set()
+    tri_vert_data = np.zeros(0, dtype=tri_vert_struct)
```
gonuke, line 349:
```suggestion
    tri_vert_data = np.zeros(len(all_tris)*3, dtype=tri_vert_struct)
```

dagmc_stats/dagmc_stats.py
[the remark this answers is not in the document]
gonuke: There are two ways to get the index:
1. start a counter before the first loop: `tri_vert_index = 0` and increment it after setting each row
2. use the index flattening with `enumerate()` as show in the screenshot

dagmc_stats/dagmc_stats.py
gonuke:
```suggestion
    roughness = np.zeros(len(all_verts))
```

dagmc_stats/dagmc_stats.py
```diff
@@ -297,3 +324,153 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
...
+    outputs
+    -------
+    roughness : (list) the roughness for all surfaces in the meshset
+    """
+    tri_vert_data, all_verts = get_tri_vert_data(my_core,
+                                                    native_ranges[types.MBTRI])
+    gc_all = get_gaussian_curvature(my_core, all_verts, tri_vert_data)
+    roughness = []
+    for vert_i in all_verts:
+        roughness.append(get_lri(vert_i, gc_all, tri_vert_data, my_core))
```
gonuke, line 475: after the above change...
```suggestion
    for vert_idx, vert_i in enumerate(all_verts):
        roughness[vert_idx] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
```

yqin43: Thank you very much! By incrementing the index each time instead of flattening(j*3+i as shown in the screenshot) it now works. The order in which the tri_vert data is added affects other parts of the code.

gonuke: I'm happy that this works now @yqin43 , but I'm not sure the order matters.  This will be something that is good to confirm and note if it is true, though!

gonuke:
Thanks for all the work on the complex new metric, @yqin43 !!!

Issue #58, opened 2019-12-09
Setup CI for PR Testing


kkiesling:
CI should be implemented so that PRs can be automatically tested.

Issue #56, opened 2019-12-04
Code structure and file clean up


kkiesling:
Currently all the code (user tool and tests) exist in a single folder which is also the top main folder. For ease of developing, the code should probably be structured into a `tool/` folder that contains generate_stats.py, dagmc_stats.py, etc. and a `tests/` folder that contains all testing files.

kkiesling:
Here is an example for the suggested code structure: https://github.com/svalinn/DAGMC-viz

Issue #30, opened 2019-08-08
remove hard coded path to PyMOAB


kkiesling:
Currently, the path to PyMOAB is hardcoded into each file. Users should include the path to PyMOAB in their PYTHONPATH in their environment so it can be automatically found. Information on what to include in PATHs should go in README (#29).

gonuke:
FYI - This is a detail of the cloud-based jupyterhub environment that I set up for @ep09772

PR #60, opened 2020-03-30
Adds CI and better testing framework

kkiesling:
## Description
The current test file was not being tested in CI and it turns out that many of the tests were failing. This is an attempt to clean it up and introduce CI.

This should address issue #30, #56, and #58.

## Changes
- new code structure
- removes hard coded python path in the scripts, but the `scripts` folder needs to be added to the `PYTHONPATH` in order for it to be imported for the tests.

## Other Information
Some tests pass only because I had to comment out the assertions. Some of the function calls were incorrect and I was not sure what the tests were supposed to be doing. @yqin43 should look at those tests and make the appropriate fixes in a different PR.


kkiesling:
Okay this is ready for review now. Note: I am not trying to make a comprehensive set of tests right now, but just setup the framework.

.circleci/config.yml
```diff
@@ -0,0 +1,16 @@
+version: 2.1
+
+jobs:
+    Test:
+        docker:
+            - image: kkiesling/dagmc_stats_depends:0.2
```
kkiesling: I will eventually move this image to the Svalinn docker hub when this gets reviewed

dagmc_stats.dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    python-setuptools \
+    python-numpy \
+    python-pip \
+    python-tables \
+    python-matplotlib \
+    git \
+    cmake \
+    vim \
+    emacs \
+    nano \
```
kkiesling: I should remove the unnecessary tools

tests/test_dagmc_stats.py
```diff
@@ -0,0 +1,142 @@
...
+            my_core, entityset_ranges)
+        surfaces = my_core.get_entities_by_type_and_tag(
+            root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [2]).size()
+        assert(len(t_p_s_data) == surfaces)
+
+        # !! This does not pass the tests and I do not know what it is supposed to do
+        # needs updating !
+
+        #triangles = my_core.get_entities_by_type(root_set, types.MBTRI).size()
+        #assert(sum(t_p_s_data) == triangles)
```
kkiesling: @yqin43 - see this test and the next two tests. These assertions were failing and I was not sure the correct way you wanted them to be tested. Once this PR is merged, you or both us can take a look at these.

dagmc_stats.dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    && git checkout -b Version5.1.0 origin/Version5.1.0 \
+    && autoreconf -fi \
+    && cd .. \
+    && mkdir build \
+    && cd build \
+    && ../moab/configure --enable-shared --enable-optimize --enable-pymoab --disable-debug --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial --prefix=$HOME/opt/moab \
+    && make \
+    && make install \
+    && cd .. \
+    && rm -rf build moab
```
bam241: why is everyone building its own docker container with MOAB in It ?

bam241: we definitely need svalinn docker image )

scripts/dagmc_stats.py
```diff
@@ -294,7 +291,7 @@ def get_coarseness(my_core, meshset, entity_ranges, geom_dim):
     """
 
     coarseness = []
-    
+
```
bam241: maybe formatting change should be part of a separate PR ?

(I always to them... and @gonuke ask me for a different PR for formatting changes... this is revenge :) )

dagmc_stats.dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# set environment variables
+ENV PATH $HOME/opt/moab/bin/:$PATH
+ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
+ENV PYTHONPATH $HOME/opt/moab/lib/python2.7/site-packages/:$PYTHONPATH
+ENV PYTHONPATH /root/project/scripts:$PYTHONPATH
```
bam241: I strongly advice against this, I would suggest either:
- install your python package in the system and rely on them to run the test
- run test from project folder and do import in test such as `from scripts import dagmc_stats`

kkiesling: Right- So then do I need to setup a `setup.py` script then to install in the system? Like @piperlincoln did for DAGMC-viz?

This was the only way I could get the module to be found when in the testing directory without  doing a system install.

scripts/dagmc_stats.py
kkiesling: My text editor automatically strips trailing whitespaces upon saving. Really there should never be trailing whitespace in a PR and we should check for that. I am not making a separate PR for this. Everyone should setup their text editor to always strip trailing whitespace upon saving so we never run into this issue anymore. I will die on that hill.

kkiesling: PS- for review purposes, you can ignore whitespace changes when you view the diffs.

dagmc_stats.dockerfile
bam241: no you don't need to install, I don't think @piperlincoln setup does not work properly anyway and she did not rely on installed packages to run the test.

kkiesling:
> Thx for putting that in, nice work @kkiesling !! :)
> 
> why is this named `dagmc_stats` if it only rely on MOAB ?
> 
> I think you need better a name than `scripts`.... maybe `dagmc_stats` ?
> 
> maybe add a `readme.md` file to describe the h5m files in `test`... otherwise you will end up eventually with a test folder "à la" PyNE with a ton of files used for tests... but nobody knows what for what....

The dockerfile was named dagmc_stats just because this is the project it was made for. I agree, we should just make a MOAB-only depends docker image in Svalinn and I can start w/ that one. I tried to use the images already existing there but they don't have PyMOAB enabled. I can make new ones though.

And yes, agreed on the folder name. I was just thinking it should be renamed.

Good idea on the test folder README- I will add that too.

dagmc_stats.dockerfile
kkiesling: Do you have a good example of how do this properly?

bam241: @piperlincoln does it correctly in her PR just ignore the setup.py :)

basically with your layout you need:
- run test from `.../project/`
- in the `test.py` `import scripts.dagmc_stats` or `from scripts import dogma_stats`

kkiesling: If I run the tests from `../project` which is the default behavior on CI, it can't find any of the test geometry .h5m files. Those are located in the test folder. Also- I want the test files to be able to run on a local machine. I don't think this method allows that.

bam241: I don't understand why `pytest` can't load your package when not running in a docker container. (I tried can't figure this out...)
otherwise best practice advice to do a setup.py
https://docs.pytest.org/en/latest/goodpractices.html

kkiesling:
@bam241 or @piperlincoln - can you take a look at how I have this setup now? I don't use a setup.py script. But I still want to be able to run tests locally in the test folder and also have them found in the docker image. And so far this is the only way I have been able to do that: set the PYTHONPATH and run pytest from inside the test directory on circleci. 

I no longer have the PYTHONPATH set in the docker image for the dagmc_stats repo, but I do have it set for PyMOAB. I added the dagmc_stats pythonpath to the circleci config, but then PyMOAB wasn't found, so I had to add PyMOAB there as well.

bam241:
> @bam241 or @piperlincoln - can you take a look at how I have the circleci config file setup now? I don't use a setup.py script. But I still want to be able to run tests locally in the test folder and also have them found in the docker image. And so far this is the only way I have been able to do that: set the PYTHONPATH and run pytest from inside the test directory on circleci.
> 
> I no longer have the PYTHONPATH set in the docker image for the dagmc_stats repo, but I do have it set for PyMOAB. I added the dagmc_stats pythonpath to the circleci config, but then PyMOAB wasn't found, so I had to add PyMOAB there as well.

I am sorry to insist, but I really think that appending to the `PYTHONPATH` is not a good and clean method. You will end-up with a `PYTHONPATH` pointing to half the folders in your computer.

I would rather se a very simple setup.py:
```
from setuptools import setup, find_packages

setup(name="dagmc_stats",  packages=find_packages())
```
and if you run a simple `pip install -e . --user` from your root repo folder you will be able to "run" locally.
(this will create a tiny `dagmc-stats.egg-link` in you local PYTHON lib folder containing:
```
/path_to/dagmc_stats
.
```

this allows you to keep your PYTHONPATH clean

to remove the egg file you can: `pip uninstall dagmc-stats`

kkiesling:
Ya ok- I think I understand what you all are saying now. This whole process of managing CI w/ or w/out a PYTHONPATH and a setup.py is all new to me. Let me try again. I will make a setup.py and try again.

tests/test_dagmc_stats.py
```diff
@@ -0,0 +1,142 @@
...
+        dagmc_tags = dagmc_stats.get_dagmc_tags(my_core)
+        entityset_ranges = dagmc_stats.get_entityset_ranges(my_core, root_set, dagmc_tags['geom_dim'])
+        node_range = my_core.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [0])
+        assert(node_range == entityset_ranges['Nodes'])
+        curve_range = my_core.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [1])
+        assert(curve_range == entityset_ranges['Curves'])
+        surface_range = my_core.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [2])
+        assert(surface_range ==entityset_ranges['Surfaces'])
+        volume_range = my_core.get_entities_by_type_and_tag(root_set, types.MBENTITYSET, dagmc_tags['geom_dim'], [3])
+        assert(volume_range == entityset_ranges['Volumes'])
```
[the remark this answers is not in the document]
kkiesling: In this PR I am not addressing the actual tests and whether or not they are good. I am just trying to get the framework setup so that the tests can actually be used at all. But that is a good comment about only doing one assertion per test (unless it can't be untangled).

dagmc_stats.dockerfile
kkiesling: If I run pytest in the main dir (not the test dir) it can't find the geometry files. I just added a setup.py (need to push still) so it can find the module just fine now wherever I am. But it still can't find the geometry files in the test dir to actually run the tests.

bam241: now you should be able to run from the test folder :)

kkiesling:
okay- check the setup.py and config.yml that I have now?

I need to add a second job that tests on python 2 as well. So this isn't quite ready.

.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+    test_py3:
+        docker:
+            - image: kkiesling/pymoab-py3-18.04
+        steps:
+            - checkout
+            - run:
+                working_directory: tests/
+                command: |
+                    pip install -e .. --user
```
bam241, line 12: I would separate the test and the pip instal.
maybe just add a:
`- run: pip install -e . --user` after the `- checkout` ?

bam241:
This is looking good, 
one minor comment and I'll be ready to merge it :)

kkiesling:
Thanks for all the input @bam241 and @piperlincoln! I think this is ready now.

Also- I had no idea you could use `pip` like that to install local packages. This is a game changer for so much of what I do.

bam241:
> Thanks for all the input @bam241 and @piperlincoln! I think this is ready now.
> 
> Also- I had no idea you could use `pip` like that to install local packages. This is a game changer for so much of what I do.

I discovered it because of this. It is pretty cool :)

bam241:
lastly could you move your dockerfile into svalinn/dockerfile and push it to svalinn dockerhub ?

then change you circle.yml to point to it ?

kkiesling:
> lastly could you move your dockerfile into svalinn/dockerfile and push it to svalinn dockerhub ?
> 
> then change you circle.yml to point to it ?
> 
> the good thing is that because it is already pushed it should be super fast....

done

bam241:
thx a ton! @kkiesling 
Merging !

PR #61, opened 2020-04-14
More CI/testing updates

kkiesling:
As I am learning more and more about CI, testing, and pytest, I realized there are updates needed here. This PR does the following:
1. remove the pip install line from CI (unnecessary)
2. Changes the paths of the .h5m test files to be relative to the top level directory so that pytest can be run from the top level
3. runs pytest from the top level directory on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,18 +6,14 @@ jobs:
             - image: svalinn/pymoab-py3-18.04
         steps:
             - checkout
-            - run:  pip install -e . --user
             - run:
-                working_directory: tests/
                 command: pytest
     test_py2:
         docker:
             - image: svalinn/pymoab-py2-18.04
         steps:
             - checkout
-            - run:  pip install -e . --user
             - run:
-                working_directory: tests/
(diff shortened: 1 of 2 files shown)


kkiesling:
> I think it is fine to remove `pip install -e . --user`.
> But I would suggest to install it using pip (with out the `-e`) and try to import it from the `ROOT` folder.

I don't quite understand what you are suggesting

bam241:
basically having a job checking the proper installation of the package, and that you can import it

.circleci/config.yml
```diff
@@ -1,6 +1,24 @@
...
 
 jobs:
+    install_py3:
+        docker:
+            - image: svalinn/pymoab-py3-18.04
+        steps:
+            - checkout
+            - run:
+                command: pip install . --user
+                command: python -c "import dagmc_stats"
```
bam241, line 11: could you do that in `$root` ? not the repo folder ?

bam241:
LGTM! thx @kkiesling 

Merging on test completion

PR #66, opened 2020-10-13
Roughness

yqin43:
## Description
Added code for adding tags and code for calculating average roughness value of the geometry

## Changes
Updated get_area_triangle function so it can calculate area of a single triangle
Tagged local roughness
Added generic code for adding tags
Added code and test for average roughness value


dagmc_stats/dagmc_stats.py
```diff
@@ -275,7 +275,7 @@ def get_triangle_aspect_ratio(my_core, meshset, geom_dim):
     return t_a_r
 
 
-def get_area_triangle(my_core, meshset, geom_dim):
+def get_area_triangle(my_core, meshset, geom_dim, tris=[]):
```
kkiesling, line 278: A test needs to be added to reflect this change. Currently only no `tris` argument is tested, but one needs to be added that tests if a subset of `tris` is added.

kkiesling: Also on that test, we should be checking the values of `obs` vs `exp`, not just the length of the data returned.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +471,71 @@ def get_roughness(my_core, native_ranges):
 
     outputs
     -------
-    roughness : (numpy array) the roughness for all surfaces in the meshset
+    roughness : (dictionary) the roughness for all vertices in the meshset
```
kkiesling, line 474: specify what the keys and values should be in this dictionary.

dagmc_stats/dagmc_stats.py
```diff
@@ -460,7 +460,7 @@ def get_lri(vert_i, gc_all, tri_vert_data, my_core):
     return Lri
 
 
-def get_roughness(my_core, native_ranges):
+def get_roughness(my_core, native_ranges, tag=True):
```
kkiesling, line 463: docstring needs the `tag` input added below.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +471,71 @@ def get_roughness(my_core, native_ranges):
...
     tri_vert_data, all_verts = get_tri_vert_data(my_core,
                                                     native_ranges[types.MBTRI])
     gc_all = get_gaussian_curvature(my_core, all_verts, tri_vert_data)
-    roughness = np.zeros(len(all_verts))
-    for vert_idx, vert_i in enumerate(all_verts):
-        roughness[vert_idx] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
+    roughness = {}
+    for vert_i in all_verts:
+        roughness[vert_i] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
+    if tag:
```
kkiesling, line 482: Right now this will add the tag to the moab instance in memory but it doesn't look like it gets written to a file. I think that in `generate_stats.py` a new file should be written out if tags are added after all the stats calculations are finished. So there needs to be some way to determine if tags have been set. 

One thought is that the value `tag` can be returned from this function in `generate_stats.py`. If after all stats have been run, if `tag` is true, then write the file. Otherwise we don't need to write out a new file. This will allow other tags to be added to the geometry as well and just write out a file with all tags.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +471,71 @@ def get_roughness(my_core, native_ranges):
...
+            three_verts = list(my_core.get_adjacencies(tri, 0, op_type=1))
+            sum_lr = 0
+            for vert in three_verts:
+                sum_lr += roughness[vert]
+            lr_tag[tri] = sum_lr/3
+        add_tag(my_core, 'FACET_AVG_LR', lr_tag, types.MB_TYPE_DOUBLE)
     return roughness
+    
+    
+def add_tag(my_core, tag_name, tag_dic, tag_type):
```
kkiesling, line 495: this function needs a test- we can discuss if you don't have an idea for how to test this.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +471,71 @@ def get_roughness(my_core, native_ranges):
...
+    tag_name : tag name
+    tag_dic : a dictionary containing eh:data
+    tag_type : tag type
+    """
+    # create the tag handle
+    tag_eh = \
+    my_core.tag_get_handle(tag_name, size=1,
+                            tag_type=tag_type,
+                            storage_type=types.MB_TAG_SPARSE,
+                            create_if_missing=True)
```
kkiesling, line 510:
```suggestion
    tag_eh = \
        my_core.tag_get_handle(tag_name, size=1,
                                tag_type=tag_type,
                                storage_type=types.MB_TAG_SPARSE,
                                create_if_missing=True)
```
PEP8

dagmc_stats/dagmc_stats.py
kkiesling: I also think the default should be False

dagmc_stats/dagmc_stats.py
kkiesling, line 482:
```suggestion
    
    if tag:
```
recommended blank line for better readability

dagmc_stats/dagmc_stats.py
kkiesling: Also kudos on this function- This will be a good one for reusing in many contexts.

kkiesling (changes requested):
great work! Overall the additions are great. I don't have any major suggestions on what you've added, but I think there are a few things missing still such as some tests and a way to write out a newly tagged geometry.

kkiesling:
Also- python3 test is failing so we need to look into that.

tests/test_dagmc_stats.py
```diff
@@ -8,7 +8,7 @@
 import unittest
 
 entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
-test_env = [ { 'input_file' : 'tests/3vols.h5m'}, {'input_file' : 'tests/single-cube.h5m'} ]
+test_env = [ { 'input_file' : 'tests/3vols.h5m'}, {'input_file' : 'tests/single-cube.h5m'}, {'input_file' : 'tests/pyramid-2.h5m'}]
```
kkiesling, line 11: the `test_get_lri()` function should use this new test geometry. I don't think it makes sense for us to use the cube when all the values are zero (I think that is also why the test is failing in python 3).

tests/test_dagmc_stats.py
```diff
@@ -213,10 +213,53 @@ def test_get_roughness(self):
...
+        """
+        # single_cube
+        my_core = test_env[1]['core']
+        native_ranges = test_env[1]['native_ranges']
+        roughness = ds.get_roughness(my_core, native_ranges)
+        geom_dim = test_env[1]['dagmc_tags']['geom_dim']
+        
+        exp = 0
+        obs = ds.avg_roughness(my_core, roughness, geom_dim)
+        self.assertAlmostEqual(exp, obs)
```
kkiesling, line 236: let's just do the pyramid not the cube for any roughness tests. (see above comment).

tests/test_dagmc_stats.py
kkiesling: also this file can just be called `pyramid.h5m` since there is not a pyramid-1 anywhere.

tests/test_dagmc_stats.py
```diff
@@ -213,10 +213,53 @@ def test_get_roughness(self):
...
+        my_core = test_env[2]['core']
+        native_ranges = test_env[2]['native_ranges']
+        roughness = ds.get_roughness(my_core, native_ranges)
+        geom_dim = test_env[2]['dagmc_tags']['geom_dim']
+        
+        gc_top = 2.0/3*np.pi
+        gc_bottom = 5.0/6*np.pi
+        d_bottom = [1/np.tan(np.pi/3),
+                    0.5*(1/np.tan(np.pi/3)+1/np.tan(np.pi/4)), 0]
+        lr_bottom=[]
```
kkiesling, line 248:
```suggestion
        lr_bottom = []
```
pep8

tests/test_dagmc_stats.py
```diff
@@ -213,10 +213,53 @@ def test_get_roughness(self):
...
+                    /(d_bottom[0]+2*d_bottom[1])))
+        lr_bottom.append(np.abs(gc_bottom - \
+                    (d_bottom[0]*gc_top+2*d_bottom[1]*gc_bottom+d_bottom[2]*gc_bottom) \
+                    /(d_bottom[0]+2*d_bottom[1]+d_bottom[2])))
+        s_top = 4*25/4*np.sqrt(3)
+        s_bottom = [12.5+2*25/4*np.sqrt(3), 25+2*25/4*np.sqrt(3)]
+        num = lr_top*s_top/3 + \
+                    2*lr_bottom[0]*s_bottom[0]/3 + \
+                    2*lr_bottom[1]*s_bottom[1]/3
+        denom = 4*25/4*np.sqrt(3)+25
```
kkiesling, line 261: there are a lot of magic numbers in this section. I am mostly wondering what the 25 number is here?

kkiesling: some comments for what each of these calculations is doing will be helpful to future developers in the event that this test fails.

kkiesling (changes requested):
second review- some notes on your roughness test.

tests/test_dagmc_stats.py
yqin43: 25 is the area of the bottom surface which is a 5*5 square

dagmc_stats/dagmc_stats.py
```diff
@@ -460,23 +463,85 @@ def get_lri(vert_i, gc_all, tri_vert_data, my_core):
...
+    if tag:
+        #add triangle average roughness tag
+        lr_tag = {}
+        for tri in native_ranges[types.MBTRI]:
+            three_verts = list(my_core.get_adjacencies(tri, 0, op_type=1))
+            sum_lr = 0
+            for vert in three_verts:
+                sum_lr += roughness[vert]
+            lr_tag[tri] = sum_lr/3
+        add_tag(my_core, 'FACET_AVG_LR', lr_tag, types.MB_TYPE_DOUBLE)
```
gonuke, line 497: Separation of concerns: Don't call `add_tag` from within this method.  Allow the user to calculate the roughness of vertices and roughness of triangles without changing the mesh DB (adding a tag changes the mesh DB).  By saving this data, the user can then call `add_tag` separately in the method that calls this.

dagmc_stats/dagmc_stats.py
```diff
@@ -460,23 +463,85 @@ def get_lri(vert_i, gc_all, tri_vert_data, my_core):
...
-    roughness = np.zeros(len(all_verts))
-    for vert_idx, vert_i in enumerate(all_verts):
-        roughness[vert_idx] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
+    roughness = {}
+    for vert_i in all_verts:
+        roughness[vert_i] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
+
+    if tag:
+        #add triangle average roughness tag
+        lr_tag = {}
```
gonuke, line 490: Create a separate method that will evaluate roughness for triangles, assuming that the vertex roughness data already exists.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +474,88 @@ def get_roughness(my_core, native_ranges):
...
+    ------
+    my_core : a MOAB Core instance
+    native_ranges : a dictionary containing ranges for each native type
+    in the file (VERTEX, TRIANGLE, ENTITYSET)
+    roughness : (dictionary) the roughness for all vertices in the meshset
+    stored in the form of vert : local roughness value
+
+    outputs
+    -------
+    tri_roughness : (dictionary) the avergae roughness values for all triangles
```
gonuke, line 502:
```suggestion
    tri_roughness : (dictionary) the average roughness values for all triangles
```

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +474,88 @@ def get_roughness(my_core, native_ranges):
...
-    roughness = np.zeros(len(all_verts))
-    for vert_idx, vert_i in enumerate(all_verts):
-        roughness[vert_idx] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
+    roughness = {}
+    for vert_i in all_verts:
+        roughness[vert_i] = get_lri(vert_i, gc_all, tri_vert_data, my_core)
     return roughness
+    
+
+def get_tri_roughness(my_core, native_ranges, roughness):
```
gonuke, line 489: Might be useful if this allowed the user to pass in a list of triangles, like `get_area_triangle()` above.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +474,88 @@ def get_roughness(my_core, native_ranges):
...
+        my_core.tag_get_handle(tag_name, size=1,
+                                tag_type=tag_type,
+                                storage_type=types.MB_TAG_SPARSE,
+                                create_if_missing=True)
+    for eh, data in tag_dic.items():
+        # assign data to the tag:
+        my_core.tag_set_data(tag_eh, eh, data)
+
+
+def avg_roughness(my_core, roughness, geom_dim):
```
gonuke, line 536: Allow the user to pass in a list of vertices, and use all the vertices only if that list is empty.

dagmc_stats/dagmc_stats.py
```diff
@@ -471,12 +474,88 @@ def get_roughness(my_core, native_ranges):
...
+    avg_roughness : the average roughness value for the geometry
+    """
+    root_set = my_core.get_root_set()
+    area_sum = sum(get_area_triangle(my_core, root_set, geom_dim))
+    num = 0
+    for vert in roughness:
+        adj_tris = my_core.get_adjacencies(vert, 2, op_type=0)
+        for tri in adj_tris:
+            s_i = get_area_triangle(my_core, root_set, geom_dim, tris=[tri])
+            num += (roughness[vert] * s_i[0] / 3)
```
gonuke, line 559:
```suggestion
        s_i = get_area_triangle(my_core, root_set, geom_dim, tris=adj_tris)
        num += roughness[vert] * sum(s_i) / 3
```

gonuke (changes requested):
A few more suggestions

dagmc_stats/dagmc_stats.py
gonuke: Actually - let's wait on this until we move to Pandas, because at some point we probably want to just populate this for the whole set of vertices

dagmc_stats/dagmc_stats.py
yqin43: Thank you for the suggestion. I made the updates in the get_roughness function instead, so the roughness value can be that of all the vertices in the geometry or the list of vertices that user passes in.

gonuke (approved):

kkiesling (approved):
approving so this can be merged

gonuke:
Decided to merge this as is.  The methods are functioning as intended, and the interfaces will be changing anyway as we introduce the class structure.

PR #68, opened 2020-10-15
Add Docker Hub Authentication

connoramoreno:
Add authorization block for Docker image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,11 @@ jobs:
 workflows:
     build:
         jobs:
-            - install_py3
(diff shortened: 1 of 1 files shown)


bam241:
thx @connoramoreno 
I think this is missing the context for each job in the workflow section
@ypark234 do you know why this is working without the context ?

kkiesling:
Thanks @connoramoreno.

For future reference, PRs should come from your fork of the repo rather than a branch on the upstream (in this case, Svalinn) repo. It's probably not worth fixing that for this PR, but FYI.

PR #70, opened 2021-02-04
Pandas

yqin43:
## Description
Pandas data frame is used to handle data instead of the dictionaries/lists used before.

## Motivation and Context
Pandas data frame supports more functions compared with dictionaries, lists and etc. It enables more consistent data storage and management. Also, it improves code efficiency.


gonuke (changes requested):
This is a great start and will definitely help us move forward quickly

PR #77, opened 2021-03-04
Dagmc stats

yqin43:
## Description
DagmcStats.py file is added to handle data populating and query problems.
test_ds.py file is added to test DagmcStats.py file.


gonuke:
It looks like we need to add `pandas` to the docker image being used for CI

gonuke:
I have made the changes to the dockerfiles and when that PR is merged we should get new dockerfiles for CI

tests/test_ds.py
```diff
@@ -0,0 +1,30 @@
...
+
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+
+    print(single_cube.root_set)
+    assert(single_cube.root_set == 0)
+
+    print(single_cube._vert_data)
+    print(single_cube._tri_data)
+    print(single_cube._surf_data)
+    print(single_cube._vol_data)
```
kkiesling, line 21: Rather than printing this information to the screen, we should be checking that this data is what we expect it to be and do something like `assert(single_cube._vert_data == [what we expect it to be])`

tests/test_ds.py
```diff
@@ -0,0 +1,30 @@
...
+    pass
```
kkiesling, line 22: `pass` isn't needed if there is a test written.

kkiesling (changes requested):
The constructor looks great. I have no real comments on the class constructor itself. I have some comments on the tests though. Let me know if you have questions on the tests.

And as @gonuke mentioned, we should remove the changes from the `generate_stats.py` file, so that can be done with these commands:

```
git checkout main -- generate_stats.py
git commit -m "revert generate_stats.py to previous version"
```

kkiesling:
> I have made the changes to the dockerfiles and when that PR is merged we should get new dockerfiles for CI

Dockerfiles should be updated

gonuke:
The docker images are updated, but tests are now failing because your test can't find the `h5m` files that they need.  I think this is because of your directory structure and where you have the files stored.

tests/test_ds.py
yqin43: I would like to discuss this during the meeting. Thanks!

tests/test_ds.py
```diff
@@ -0,0 +1,43 @@
...
+    'input_file': 'tests/single-cube.h5m'}, {'input_file': 'tests/pyramid.h5m'}]
+
+
+def test_load_file():
+    """Tests loading file and check the values of all the variables that are set in the constructor
+    """
+
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+
+    print(single_cube.root_set)
```
kkiesling, line 16: This print statement should be removed.

tests/test_ds.py
```diff
@@ -0,0 +1,30 @@
...
+
+test_env = [{'input_file': '3vols.h5m'}, {
+    'input_file': 'single-cube.h5m'}, {'input_file': 'pyramid.h5m'}]
+
+
+def test_load_file():
+    """Tests loading file and check the values of all the variables that are set in the constructor
+    """
+
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
```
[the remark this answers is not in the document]
kkiesling: Were you planning to change the structure of `test_env` to be a dictionary of dictionaries? Or leave it as is. I am fine with either way you choose, but just need to know if we are waiting on a change here.

kkiesling (commented):
excellent. There is still a print statement remaining and then it should be about ready.

yqin43:
@kkiesling Thanks! I personally would prefer to leave the structure of test_env as it is now.

gonuke (approved):
Looks good to me - thanks @yqin43 

I'll let @kkiesling merge once she clears her review

kkiesling (approved):
Excellent! Thanks @yqin43 !

PR #78, opened 2021-03-16
get_native_ranges

yqin43:
## Description
Added the get_native_ranges method and its test.

### DIFF
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -40,3 +40,23 @@ class DagmcStats:
         self._vol_data = pd.DataFrame(
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
+
+    def get_native_ranges(self, meshset, entity_types):
+        """Get a dictionary with MOAB ranges for each of the requested entity types
+
+        inputs
+        ------
+        meshset : a MOAB meshset to query for the ranges of entities
+        entity_types : a list of valid pyMOAB types to be retrieved
+
+        outputs
+        -------
+        native_ranges : a dictionary with one entry for each entity type that is a
+                        Range of handles to that type
+        """
+
+        native_ranges = {}
+        for entity_type in entity_types:
+            native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
+                meshset, entity_type)
+        return native_ranges
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -5,9 +5,8 @@ from pymoab import core, types
 
 
 class DagmcStats:
-    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
-    def __init__(self, filename, populate=False, entity=entity_types):
+    def __init__(self, filename, populate=False):
         """Constructor
 
         inputs
@@ -41,9 +40,11 @@ class DagmcStats:
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
         
-        self.native_ranges = self.__set_native_ranges(entity)
+        self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+        self.native_ranges = {}
+        self.__set_native_ranges()
 
-    def __set_native_ranges(self, entity_types):
+    def __set_native_ranges(self):
         """Get a dictionary with MOAB ranges for each of the requested entity types
 
         inputs
@@ -55,9 +56,6 @@ class DagmcStats:
         native_ranges : a dictionary with one entry for each entity type that is a
                         Range of handles to that type
         """
-
-        native_ranges = {}
-        for entity_type in entity_types:
-            native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
+        for entity_type in self.entity_types:
+            self.native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
                 self.root_set, entity_type)
-        return native_ranges
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -45,11 +45,12 @@ class DagmcStats:
         self.__set_native_ranges()
 
     def __set_native_ranges(self):
-        """Get a dictionary with MOAB ranges for each of the requested entity types
+        """Set the class native_ranges variable to a dictionary with MOAB
+        ranges for each of the requested entity types
 
         inputs
         ------
-        entity_types : a list of valid pyMOAB types to be retrieved
+        none
 
         outputs
         -------
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -5,6 +5,7 @@ from pymoab import core, types
 
 
 class DagmcStats:
+    entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
     def __init__(self, filename, populate=False):
         """Constructor
@@ -22,7 +23,6 @@ class DagmcStats:
         self._my_moab_core = core.Core()
         self._my_moab_core.load_file(filename)
         self.root_set = self._my_moab_core.get_root_set()
-        entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
         # initialize data frames
         self._vert_data = pd.DataFrame(
@@ -41,12 +41,11 @@ class DagmcStats:
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
 
-    def get_native_ranges(self, meshset, entity_types):
+    def __set_native_ranges(self, entity_types):
         """Get a dictionary with MOAB ranges for each of the requested entity types
 
         inputs
         ------
-        meshset : a MOAB meshset to query for the ranges of entities
         entity_types : a list of valid pyMOAB types to be retrieved
 
         outputs
@@ -55,8 +54,8 @@ class DagmcStats:
                         Range of handles to that type
         """
 
-        native_ranges = {}
+        self.native_ranges = {}
         for entity_type in entity_types:
-            native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
-                meshset, entity_type)
-        return native_ranges
+            self.native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
+                self.root_set, entity_type)
+        return self.native_ranges
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -7,7 +7,7 @@ from pymoab import core, types
 class DagmcStats:
     entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
-    def __init__(self, filename, populate=False):
+    def __init__(self, filename, populate=False, entity=entity_types):
         """Constructor
 
         inputs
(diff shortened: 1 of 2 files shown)


tests/test_ds.py
```diff
@@ -3,6 +3,7 @@
 import dagmc_stats.DagmcStats as ds
 import pandas as pd
 
+entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
```
gonuke, line 6: We should make `entity_types` a class variable of `DagmcStats` and then we can refer to it in the tests as `ds.entity_types`

tests/test_ds.py
```diff
@@ -13,29 +14,47 @@ def test_load_file():
...
+def test_native_ranges():
+    """Tests get_native_ranges
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    meshset = single_cube.root_set
+    native_ranges = single_cube.get_native_ranges(meshset, entity_types)
+
+    vertex_range = single_cube._my_moab_core.get_entities_by_type(
+        meshset, types.MBVERTEX)
+    assert(vertex_range == native_ranges[0])
```
gonuke, line 54: Should use the current type as the argument to `native_ranges[]`

Also, perhaps this can be a loop over `entity_types`?

gonuke (changes requested):
This is great incremental work - easy to review and get merged.

I have a couple of small suggestions

tests/test_ds.py
gonuke:
```suggestion
    assert(vertex_range == native_ranges[types.MBVERTEX])
```

dagmc_stats/DagmcStats.py
```diff
@@ -40,3 +40,23 @@ def __init__(self, filename, populate=False):
         self._vol_data = pd.DataFrame(
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
+
+    def get_native_ranges(self, meshset, entity_types):
```
gonuke, line 44: If `entity_types` becomes a class variable.
```suggestion
    def get_native_ranges(self, meshset, entity_types = DagmcStats.entity_types):
```

dagmc_stats/DagmcStats.py
```diff
@@ -40,3 +40,23 @@ def __init__(self, filename, populate=False):
...
+        meshset : a MOAB meshset to query for the ranges of entities
+        entity_types : a list of valid pyMOAB types to be retrieved
+
+        outputs
+        -------
+        native_ranges : a dictionary with one entry for each entity type that is a
+                        Range of handles to that type
+        """
+
+        native_ranges = {}
```
gonuke, line 58: Directly set the class member variable instead of returning it
```suggestion
        self.native_ranges = {}
```

dagmc_stats/DagmcStats.py
gonuke: Also change this to called `set_native_ranges()`

gonuke: Final (?) suggestion
```suggestion
    def _set_native_ranges(self):
```

kkiesling, line 44: let's rework this as a "convenience function" just for the sake of setting this information in the constructor:

```
def __set_native_ranges(self):
```

The meshset is just `self.root_set`. And the entity types are the list you are originally passing in, but we don't need that as a separate passed variable and can have that set in the function directly.

Making it private with the double underscore indicates that this is only to be used as a helper function in the constructor and not as an accessible function to users.

tests/test_ds.py
```diff
@@ -13,29 +14,47 @@ def test_load_file():
...
     # another test to test load_file() with populate=True
     pass
+
+
+def test_native_ranges():
+    """Tests get_native_ranges
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    meshset = single_cube.root_set
+    native_ranges = single_cube.get_native_ranges(meshset, entity_types)
```
kkiesling, line 50: if you make this a private function `__set_native_ranges()`, then to call the private function is a bit more complicated:

`single_cube._DagmcStats__set_native_ranges()`

tests/test_ds.py
kkiesling: let's also construct the expected dictionary and test that it is equivalent to the `native_ranges` output.

kkiesling: so that means just constructing one dictionary and doing one assertion at the end of the file rather than the three separate asserts:

`assert(sorted(exp_dict) == sorted(native_ranges))`

dagmc_stats/DagmcStats.py
```diff
@@ -7,7 +7,7 @@
 class DagmcStats:
     entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
 
-    def __init__(self, filename, populate=False):
+    def __init__(self, filename, populate=False, entity=entity_types):
```
gonuke, line 10: I don't foresee ever calling the constructor with a different set of entity types? Is there a particular use case imagined here?

dagmc_stats/DagmcStats.py
```diff
@@ -40,6 +40,8 @@ def __init__(self, filename, populate=False):
         self._vol_data = pd.DataFrame(
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
+        
+        self.native_ranges = self.__set_native_ranges(entity)
```
gonuke, line 44: Do we need to pass in the entity types, or just use the ones defined as a class variable?

dagmc_stats/DagmcStats.py
```diff
@@ -54,8 +56,8 @@ def __set_native_ranges(self, entity_types):
...
         """
 
-        self.native_ranges = {}
+        native_ranges = {}
         for entity_type in entity_types:
-            self.native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
+            native_ranges[entity_type] = self._my_moab_core.get_entities_by_type(
                 self.root_set, entity_type)
-        return self.native_ranges
+        return native_ranges
```
gonuke, line 63: No need to return anything - this is now a "set" method rather than a "get" method

gonuke (changes requested):
Thanks for the updates @yqin43 - I have a few comments that you can discuss with @kkiesling this morning....

dagmc_stats/DagmcStats.py
yqin43: I think this is the way to access the class variable in __init__() function, otherwise it will cause error.

tests/test_ds.py
```diff
@@ -13,29 +14,42 @@ def test_load_file():
...
+    """Tests get_native_ranges
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    meshset = single_cube.root_set
+    native_ranges = single_cube._DagmcStats__set_native_ranges(entity_types)
+    
+    for native_range_type in single_cube.entity_types:
+        range = single_cube._my_moab_core.get_entities_by_type(
+            meshset, native_range_type)
+        assert(range == native_ranges[native_range_type])
```
kkiesling, line 55: assert a Truth array outside the for-loop so that we can test all three results even if one of them is wrong. If one doesn't pass the assertion before it gets to the others, it won't run the rest of them to check.

dagmc_stats/DagmcStats.py
```diff
@@ -22,7 +23,6 @@ def __init__(self, filename, populate=False):
         self._my_moab_core = core.Core()
         self._my_moab_core.load_file(filename)
         self.root_set = self._my_moab_core.get_root_set()
-        entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
```
kkiesling, line 25: If we make this `self.entity_types = ...` then it doesn't need to be passed the constructor above

dagmc_stats/DagmcStats.py
kkiesling: don't need to pass `entity` here if we are using `self.entity_types` as a class variable.

tests/test_ds.py
```diff
@@ -13,29 +14,42 @@ def test_load_file():
...
     # another test to test load_file() with populate=True
     pass
+
+
+def test_native_ranges():
+    """Tests get_native_ranges
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    meshset = single_cube.root_set
+    native_ranges = single_cube._DagmcStats__set_native_ranges(entity_types)
```
kkiesling, line 50:
```suggestion
    native_ranges = single_cube._DagmcStats__set_native_ranges()
```
for consistency with changing to a class variable

kkiesling (changes requested):

tests/test_ds.py
kkiesling: actually, since this is now called in the constructor we don't need to explicitly call it here and can just test that it is set properly in the constructor.

dagmc_stats/DagmcStats.py
```diff
@@ -40,3 +39,23 @@ def __init__(self, filename, populate=False):
...
+        self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+        self.native_ranges = {}
+        self.__set_native_ranges()
+
+    def __set_native_ranges(self):
+        """Get a dictionary with MOAB ranges for each of the requested entity types
+
+        inputs
+        ------
+        entity_types : a list of valid pyMOAB types to be retrieved
```
kkiesling, line 52: no inputs anymore

kkiesling (approved):
Looks ready to me- I will let @gonuke approve and merge

dagmc_stats/DagmcStats.py
```diff
@@ -40,3 +39,24 @@ def __init__(self, filename, populate=False):
...
+        inputs
+        ------
+        none
+
+        outputs
+        -------
+        native_ranges : a dictionary with one entry for each entity type that is a
+                        Range of handles to that type
+        """
+        for entity_type in self.entity_types:
```
gonuke, line 60: It may make more sense to explicitly refer to the class variable here?  It appears that both will work, but `DagmcStats.entity_types` is accessible in various contexts/scopes, and it may be more logical to use it that way throughout.

```suggestion
        for entity_type in DagmcStats.entity_types:
```

dagmc_stats/DagmcStats.py
```diff
@@ -40,3 +39,24 @@ def __init__(self, filename, populate=False):
         self._vol_data = pd.DataFrame(
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
         self._vol_data = self._vol_data.set_index('vol_eh')
+        
+        self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
```
gonuke, line 43: I think this still makes more sense as a class variable, and it can be accessed directly in the member function, as described below.  That said, you may have discussed a reason that it should not be a class variable...?

tests/test_ds.py
```diff
@@ -13,29 +14,41 @@ def test_load_file():
...
     # another test to test load_file() with populate=True
     pass
+
+
+def test_native_ranges():
+    """Tests get_native_ranges
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    test_pass = np.full(3, False)
+    for i, native_range_type in enumerate(single_cube.entity_types):
```
gonuke, line 50: If this was reverted to a class variable, then this would become:
```suggestion
    for i, native_range_type in enumerate(ds.entity_types):
```

gonuke (changes requested):
I don't have the benefit of your conversation this morning, but I think there is a path forward in which `entity_types` is still a class variable, unless there is a reason not to.

tests/test_ds.py
kkiesling: no- the object is created in line 48 above. It is already a class variable.

dagmc_stats/DagmcStats.py
kkiesling: I don't understand what you are asking for. Is this not a class variable being set in the init?

kkiesling:
Perhaps we/I misunderstood what you were recommending wrt a class variable @gonuke

gonuke (approved):

gonuke:
This works fine this way, and it won't matter for our use cases.

PR #80, opened 2021-04-13
set_entityset_ranges

yqin43:
## Description
Added set_entityset_ranges() function and its test

### DIFF
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -41,6 +41,7 @@ class DagmcStats:
         self._vol_data = self._vol_data.set_index('vol_eh')
 
         self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
+        self.entityset_types = {0:'Nodes', 1:'Curves', 2:'Surfaces', 3:'Volumes'}
         self.native_ranges = {}
         self.__set_native_ranges()
         self.dagmc_tags = {}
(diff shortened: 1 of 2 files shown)


dagmc_stats/DagmcStats.py
```diff
@@ -90,3 +92,23 @@ def __set_dagmc_tags(self):
...
+        inputs
+        ------
+        none
+
+        outputs
+        -------
+        none
+        """
+
+        entityset_types = ['Nodes', 'Curves', 'Surfaces', 'Volumes']
```
gonuke, line 110: I'd argue for this being either a Class variable or an object member variable (I still prefer the former for this and `self.entity_types`)

kkiesling, line 110: I think we need to explicitly tie the `dimension` integer to the name of the type rather than rely on it's position in the list like we do not. It's not very robust if someone comes along and changes the order of this list.

tests/test_ds.py
```diff
@@ -64,3 +64,18 @@ def test_set_dagmc_tags():
...
     assert(single_cube.dagmc_tags['global_id'])
+
+
+def test_set_entityset_ranges():
+    """
+    Tests different aspects of the set_entityset_ranges function
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    entityset_types = ['Nodes', 'Curves', 'Surfaces', 'Volumes']
+    test_pass = np.full(4, False)
```
gonuke, line 75:
```suggestion
    test_pass = np.full(len(entityset_types), False)
```

gonuke (changes requested):
The list of possible entity set types should probably be a class variable

kkiesling (approved):
This looks good now. I will let @gonuke confirm the changes too.

gonuke (approved):

PR #81, opened 2021-04-20
get_tris

yqin43:
## Description
get_tris() function
Test is not added so this is only a draft pull request

### DIFF
--- a/dagmc_stats/DagmcStats.py
+++ b/dagmc_stats/DagmcStats.py
@@ -95,8 +95,7 @@ class DagmcStats:
                                                                      create_if_missing=False)
 
     def __set_entityset_ranges(self):
-        """
-        Set a dictionary with MOAB Ranges that are specific to the
+        """Set a dictionary with MOAB Ranges that are specific to the
         types.MBENTITYSET type
 
         inputs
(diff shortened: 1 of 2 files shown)


dagmc_stats/DagmcStats.py
```diff
@@ -111,3 +110,40 @@ def __set_entityset_ranges(self):
...
                                                                 self.dagmc_tags['geom_dim'], [dimension])
+
+    def get_tris(self, meshset=None):
+        """Get triangles of a volume if geom_dim is 3
+        Get triangles of a surface if geom_dim is 2
+        Else get all the triangles
+
+        inputs
+        ------
+        meshset : set of entities that are used to populate data. Default value
```
kkiesling, line 121: @gonuke - question for you. @yqin43 and I were thinking that it might be useful if the user could pass `meshset` as a list of entity handles rather than just a single. Right now if the user happens to pass a list, it will go to line 143 and default to the `root_set`. Do you think it's useful to make this so it could be a list of different surfaces and volumes? or just make the user supply one meshset at a time?

gonuke: It might be convenient, but the logic might get pretty messy trying to sort out if each thing in the list is a surface or volume or other....  That might need a separate method to take a list of entitysets and return a set of entitysets that are only surfaces...?

kkiesling: Yes that was my thought. Might be convenient but also really messy. I think for now we'll leave it as is and keep this in mind if we have a use for passing in a list in the future.

gonuke (approved):
LGTM - thanks @yqin43

kkiesling (approved):

PR #82, opened 2021-05-06
DagmcStats Refactor

yqin43:
## Description
DagmcStats -> DagmcFile (rename)
Add DagmcQuery class init (separate file)
Move get_tris() from DagmcFile to DagmcQuery


dagmc_stats/DagmcStats.py
```diff
@@ -147,3 +143,69 @@ def get_tris(self, meshset=None):
...
+        for tri in tris:
+            if tri not in self._tri_data.tri_eh:
+                tri_data_row = {'tri_eh': tri}
+                side_lengths = list(self.get_tri_side_length(tri).values())
+                s = .5*(sum(side_lengths))
+                p = np.prod(s - side_lengths)
+                # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
+                tri_data_row['area'] = np.sqrt(s * p)
+                tri_data_row['aspect_ratio'] = np.prod(side_lengths) / (8 * p)
+                self._tri_data = self._tri_data.append(tri_data_row, ignore_index=True)
```
kkiesling, line 170: append this to a list of dictionaries and fill the data frame all at once when all data is calculated. `tri_data.append(tri_data_row)`

dagmc_stats/DagmcStats.py
```diff
@@ -147,3 +143,69 @@ def get_tris(self, meshset=None):
...
+        ------
+        meshset : set of entities that are used to populate data. By default,
+                  the root set will be used and data of the whole geometry will
+                  be populated.
+
+        outputs
+        -------
+        none
+        """
+        tris = self.get_tris(meshset)
```
kkiesling, line 160: can initialize a list here `tri_data`

dagmc_stats/DagmcStats.py
```diff
@@ -147,3 +143,69 @@ def get_tris(self, meshset=None):
...
+            if tri not in self._tri_data.tri_eh:
+                tri_data_row = {'tri_eh': tri}
+                side_lengths = list(self.get_tri_side_length(tri).values())
+                s = .5*(sum(side_lengths))
+                p = np.prod(s - side_lengths)
+                # sqrt(s(s - a)(s - b)(s - c)), where s = (a + b + c)/2
+                tri_data_row['area'] = np.sqrt(s * p)
+                tri_data_row['aspect_ratio'] = np.prod(side_lengths) / (8 * p)
+                self._tri_data = self._tri_data.append(tri_data_row, ignore_index=True)
+        return
```
kkiesling, line 171: return the list of dictionaries to populate the data frame in the init

dagmc_stats/DagmcStats.py
```diff
@@ -48,6 +41,9 @@ def __init__(self, filename, populate=False):
         self.__set_dagmc_tags()
         self.entityset_ranges = {}
         self.__set_entityset_ranges()
+        
+        if populate is True:
+            self.__populate_triangle_data(meshset)
```
kkiesling, line 46: move the creation of the data frame after this line so it can be populated all at once and save time

kkiesling (changes requested):

dagmc_stats/DagmcStats.py
kkiesling: so you'll have single list filled with dictionaries where each dictionary is one row

dagmc_stats/DagmcStats.py
```diff
@@ -6,7 +6,7 @@
 
 class DagmcStats:
 
-    def __init__(self, filename, populate=False):
+    def __init__(self, filename, populate=False, meshset=None):
```
kkiesling, line 9: let's not pass in a meshset- we don't load the file yet so it doesn't make sense to be able to pass this.

kkiesling (changes requested):

kkiesling (commented):
We need to have a larger discussion on the code lay out/design and how we want users to be calling this code I think before we can truly move forward.

gonuke:
It might be helpful to have a very short python script that shows how these classes will be used so that we make sure we have the same understanding.

dagmc_stats/DagmcQuery.py
```diff
@@ -0,0 +1,45 @@
+from pymoab.rng import Range
+from pymoab import core, types
+import warnings
+
+class DagmcQuery:
+    def __init__(self, dagmc_file, meshset=None):
```
kkiesling, line 6: init needs a doc string to describe the inputs and the purpose of the class

kkiesling (changes requested):
we should do two separate test files for each of the two new files `test_dagmcfile.py` and `test_dagmcquery.py`. Then we test the init for each class and the methods that exist in each file. So the Dagmc Query class needs the init tested and the `get_tri` tests should go in that same test file.

kkiesling:
> It might be helpful to have a very short python script that shows how these classes will be used so that we make sure we have the same understanding.

a short script for us or users? If the former, there are notes in the shared document that show the workflow. If the latter, I think that should come in another PR that gives examples.

gonuke:
> > It might be helpful to have a very short python script that shows how these classes will be used so that we make sure we have the same understanding.
> 
> a short script for us or users? If the former, there are notes in the shared document that show the workflow. If the latter, I think that should come in another PR that gives examples.

I as thinking of a minimal implementation for us.

kkiesling:
> > > It might be helpful to have a very short python script that shows how these classes will be used so that we make sure we have the same understanding.
> > 
> > 
> > a short script for us or users? If the former, there are notes in the shared document that show the workflow. If the latter, I think that should come in another PR that gives examples.
> 
> I as thinking of a minimal implementation for us.

Let's make a gist with the script..

kkiesling (approved):
this looks good now! Merging.

kkiesling:
> > > > It might be helpful to have a very short python script that shows how these classes will be used so that we make sure we have the same understanding.
> > > 
> > > 
> > > a short script for us or users? If the former, there are notes in the shared document that show the workflow. If the latter, I think that should come in another PR that gives examples.
> > 
> > 
> > I as thinking of a minimal implementation for us.
> 
> Let's make a gist with the script..

NVM - it's not easy to collaborate on a gist it seems. We'll add this as a script in this repo.

PR #83, opened 2021-06-16
Dagmc stats

yqin43:
## Description
Get_meshset_by_id() function and its corresponding tests.

### DIFF
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -156,7 +156,6 @@ class DagmcStats:
             warnings.warn(
                 'ID is not in the given dimension range! ' +\
                     'Empty list will be returned.')
-            meshset = []
         return meshset
 
 
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -141,11 +141,12 @@ class DagmcStats:
         if len(ids) == 0:
             return self.entityset_ranges[dim]
 
-        all_global_id = []
+        dim_ms = self._my_moab_core.create_meshset()
+        self._my_moab_core.add_entity(dim_ms, self.entityset_ranges[dim])
+        meshset = []
         for id in ids:
-            all_global_id.extend(list(self._my_moab_core.get_entities_by_type_and_tag(self.root_set,
-                                                                                      types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
-        meshset = list(set(self.entityset_ranges[dim]) & set(all_global_id))
+            meshset.extend(list(self._my_moab_core.get_entities_by_type_and_tag(dim_ms,
+                                            types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
         # if id is not in the given dim range
         if not meshset:
             warnings.warn(
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -4,6 +4,7 @@ from pymoab.rng import Range
 from pymoab import core, types
 import warnings
 
+
 class DagmcStats:
 
     def __init__(self, filename, populate=False):
@@ -34,15 +35,16 @@ class DagmcStats:
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
 
         self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
-        self.entityset_types = {0:'Nodes', 1:'Curves', 2:'Surfaces', 3:'Volumes'}
+        self.entityset_types = {0: 'Nodes',
+                                1: 'Curves', 2: 'Surfaces', 3: 'Volumes'}
         self.native_ranges = {}
         self.__set_native_ranges()
         self.dagmc_tags = {}
         self.__set_dagmc_tags()
         self.entityset_ranges = {}
         self.__set_entityset_ranges()
-        
-        #if populate is True:
+
+        # if populate is True:
         #    self.__populate_triangle_data(meshset)
 
     def __set_native_ranges(self):
@@ -106,8 +108,19 @@ class DagmcStats:
             self.entityset_ranges[set_type] = \
                 self._my_moab_core.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET,
                                                                 self.dagmc_tags['geom_dim'], [dimension])
- 
+
     def get_meshset_by_id(self, dim, ids=[]):
+        """Get meshset of the geometry with specified dimension and ids
+
+        inputs
+        ------
+        dim : dimension of the meshset
+        ids : ids of the meshset
+
+        outputs
+        -------
+        meshset : meshset of the geometry with given dimension and ids
+        """
         # if no id is passed in
         if len(ids) == 0:
             return self.entityset_ranges[dim]
@@ -115,11 +128,12 @@ class DagmcStats:
         all_global_id = []
         for id in ids:
             all_global_id.extend(list(self._my_moab_core.get_entities_by_type_and_tag(self.root_set,
-                                                types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
+                                                                                      types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
         meshset = list(set(self.entityset_ranges[dim]) & set(all_global_id))
         # if id is not in the given dim range
         if not meshset:
-            warnings.warn('ID is not in the given dimension range! All entities of specified dimension will be returned.')
+            warnings.warn(
+                'ID is not in the given dimension range! All entities of specified dimension will be returned.')
             meshset = self.entityset_ranges[dim]
         return meshset
 
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -35,8 +35,8 @@ class DagmcStats:
             columns=['vol_eh', 'surf_per_vol', 'coarseness'])
 
         self.entity_types = [types.MBVERTEX, types.MBTRI, types.MBENTITYSET]
-        self.entityset_types = {0: 'Nodes',
-                                1: 'Curves', 2: 'Surfaces', 3: 'Volumes'}
+        self.entityset_types = {0: 'nodes',
+                                1: 'curves', 2: 'surfaces', 3: 'volumes'}
         self.native_ranges = {}
         self.__set_native_ranges()
         self.dagmc_tags = {}
@@ -114,13 +114,28 @@ class DagmcStats:
 
         inputs
         ------
-        dim : dimension of the meshset
-        ids : ids of the meshset
+        dim : (Integer or String) Dimension of the meshset. 0: 'node(s)',
+                                1: 'curve(s)', 2: 'surface(s)', 3: 'volume(s)'
+        ids : (Integer) Global ID(s) of the meshset
 
         outputs
         -------
-        meshset : meshset of the geometry with given dimension and ids
+        meshset : meshset of the geometry with given dimension and ids. First,
+                  dim will be checked. If dim is invalid, the root set will be
+                  returned. Then, if id is invalid (empty or not in the given
+                  dim range), all entities with the given dim will be returned.
         """
+        if isinstance(dim, int) and dim in [0, 1, 2, 3]:
+            dim = self.entityset_types[dim]
+        elif isinstance(dim, basestring) and dim.lower() in ['node', 'curve', 'surface', 'volume', 'nodes', 'curves', 'surfaces', 'volumes']:
+            dim = dim.lower()
+            if dim[-1] != 's':
+                dim = dim + 's'
+        else:
+            # invalid dim
+            warnings.warn('Invalid dim!')
+            return self.root_set
+
         # if no id is passed in
         if len(ids) == 0:
             return self.entityset_ranges[dim]
@@ -133,7 +148,8 @@ class DagmcStats:
         # if id is not in the given dim range
         if not meshset:
             warnings.warn(
-                'ID is not in the given dimension range! All entities of specified dimension will be returned.')
+                'ID is not in the given dimension range! ' +\
+                    'All entities of specified dimension will be returned.')
             meshset = self.entityset_ranges[dim]
         return meshset
 
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -125,10 +125,13 @@ class DagmcStats:
                   returned. Then, if id is invalid (empty or not in the given
                   dim range), all entities with the given dim will be returned.
         """
+        plural_names = self.entityset_types.values()
+        sing_names = [ name[:-1] for name in plural_names]
+        all_names =  plural_names + sing_names
+
         if isinstance(dim, int) and dim in self.entityset_types.keys():
             dim = self.entityset_types[dim]
-        #print(list(my_dict.keys())[list(my_dict.values()).index(112)])
-        elif type(dim) == str and dim.lower() in ['node', 'curve', 'surface', 'volume', 'nodes', 'curves', 'surfaces', 'volumes']:
+        elif type(dim) == str and dim.lower() in all_names:
             dim = dim.lower()
             if dim[-1] != 's':
                 dim = dim + 's'
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -43,6 +43,8 @@ class DagmcStats:
         self.__set_dagmc_tags()
         self.entityset_ranges = {}
         self.__set_entityset_ranges()
+        self.dim_dict = {}
+        self.__set_dimension_meshset()
 
         # if populate is True:
         #    self.__populate_triangle_data(meshset)
@@ -109,6 +111,23 @@ class DagmcStats:
                 self._my_moab_core.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET,
                                                                 self.dagmc_tags['geom_dim'], [dimension])
 
+    def __set_dimension_meshset(self):
+        """Set the class dim_dict variable to a dictionary with the
+        meshset for each dimension
+
+        inputs
+        ------
+        none
+
+        outputs
+        -------
+        none
+        """
+        for set_type, entityset_range in self.entityset_ranges.items():
+            dim_ms = self._my_moab_core.create_meshset()
+            self._my_moab_core.add_entity(dim_ms, entityset_range)
+            self.dim_dict[set_type] = dim_ms
+
     def get_meshset_by_id(self, dim, ids=[]):
         """Get meshset of the geometry with specified dimension and ids
 
@@ -145,11 +164,9 @@ class DagmcStats:
         if len(ids) == 0:
             return self.entityset_ranges[dim]
 
-        dim_ms = self._my_moab_core.create_meshset()
-        self._my_moab_core.add_entity(dim_ms, self.entityset_ranges[dim])
         meshset = []
         for id in ids:
-            meshset.extend(self._my_moab_core.get_entities_by_type_and_tag(dim_ms,
+            meshset.extend(self._my_moab_core.get_entities_by_type_and_tag(self.dim_dict[dim],
                                             types.MBENTITYSET, self.dagmc_tags['global_id'], [id]))
         # if id is not in the given dim range
         if not meshset:
--- a/dagmc_stats/DagmcFile.py
+++ b/dagmc_stats/DagmcFile.py
@@ -122,8 +122,9 @@ class DagmcStats:
         -------
         meshset : meshset of the geometry with given dimension and ids. First,
                   dim will be checked. If dim is invalid, the root set will be
-                  returned. Then, if id is invalid (empty or not in the given
-                  dim range), all entities with the given dim will be returned.
+                  returned. Then, if id is empty, all entities with the given
+                  dim will be returned. If is is not in the given dim range),
+                  an empty list will be returned.
         """
         plural_names = self.entityset_types.values()
         sing_names = [ name[:-1] for name in plural_names]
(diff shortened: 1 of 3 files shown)


dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,34 @@ def __set_entityset_ranges(self):
...
+        -------
+        meshset : meshset of the geometry with given dimension and ids
+        """
+        # if no id is passed in
+        if len(ids) == 0:
+            return self.entityset_ranges[dim]
+
+        all_global_id = []
+        for id in ids:
+            all_global_id.extend(list(self._my_moab_core.get_entities_by_type_and_tag(self.root_set,
```
gonuke, line 130: This (py)MOAB method allows you to specify multiple tags and a value for each of them.  Therefore, you could pass `[dimTag, global_id_tag]` with values `[dim, id]` (or something like that...)

gonuke (changes requested):
One suggestion to make this simpler.

dagmc_stats/DagmcFile.py
kkiesling: OHH I didn't realize that! That's clever.

kkiesling: Can you do this for a list of `id`s though? That was part of the reason we have it in a for-loop.

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,34 @@ def __set_entityset_ranges(self):
                 self._my_moab_core.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET,
                                                                 self.dagmc_tags['geom_dim'], [dimension])
 
+    def get_meshset_by_id(self, dim, ids=[]):
+        """Get meshset of the geometry with specified dimension and ids
+
+        inputs
+        ------
+        dim : dimension of the meshset
```
kkiesling, line 117: should specify if this is an integer or string and what the valid options are for `dim`.

dagmc_stats/DagmcFile.py
gonuke: The right syntax may actually be:
`[dimTag, global_id_tag]` with values `[ [dim], [id1, id2, id2] ]`
Might be worth a little digging/experimenting.

BTW, I figured this out because this is how we implement the same method in MOAB's GeomTopoTool for the same functionality

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,34 @@ def __set_entityset_ranges(self):
...
+        ids : ids of the meshset
```
kkiesling, line 118: specify that this is a global id and that it is an integer

dagmc_stats/DagmcFile.py
kkiesling: Right now it's set up to only accept `Volumes` or `Surfaces` or `Curves` or `Nodes` (case sensitive and plural). I can see a user trying to use something like `volume` or an integer (0-3), which would cause an error because you can't call those keys in the `entityset_ranges` dictionary on line 132. One suggestion to get around this is to make a few changes:

1. change `self.entityset_types` to use all lower case strings ('nodes', 'curves', 'surfaces', 'volumes'} so that you can convert any string passed into `dim` to all lowercase if it's a string: `dim.lower()`
2. in this method, the first check should be whether `dim` is an integer or a string. If it's an integer (`[0, 1, 2, 3]`) then just call the correct string from `entityset_types` to use the correct string in the rest of the method. If it's a string, convert it to `lower()` and make sure it's a valid string according to the dictionary keys. You'll probably also want a check to all both the plural or singular form of the words (eq, both `volume` and `volumes`).
3. If neither the integer nor the string for `dim` is valid, we should add a warning or error.

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,34 @@ def __set_entityset_ranges(self):
...
+
+        all_global_id = []
+        for id in ids:
+            all_global_id.extend(list(self._my_moab_core.get_entities_by_type_and_tag(self.root_set,
+                                                                                      types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
+        meshset = list(set(self.entityset_ranges[dim]) & set(all_global_id))
+        # if id is not in the given dim range
+        if not meshset:
+            warnings.warn(
+                'ID is not in the given dimension range! All entities of specified dimension will be returned.')
```
kkiesling, line 136:
```suggestion
                'ID is not in the given dimension range! ' +\
                    'All entities of specified dimension will be returned.')
```
PEP8 suggestion for the line length

tests/test_dagmc_file.py
```diff
@@ -78,6 +78,49 @@ def test_set_entityset_ranges():
...
+
+def test_get_meshset_by_id_empty_id():
+    """
+    Tests the get_meshset_by_id function given no id
+    """
+    three_vols = ds.DagmcStats(test_env[0]['input_file'])
+    exp = three_vols._my_moab_core.get_entities_by_type_and_tag(
+        three_vols.root_set, types.MBENTITYSET, three_vols.dagmc_tags['geom_dim'], [3])
+    obs = three_vols.get_meshset_by_id('Volumes', ids=[])
+    assert(obs == exp)
```
kkiesling, line 101: can you add a comment here about what the expected behavior should be if `ids=[ ]` (ie, uses rootset instead)

kkiesling (changes requested):
This looks good. I think there needs to be an added check on the `dim` variable to handle both integers and strings (and then there should be some tests added for the various cases) to be more robust and match how anticipated user use.

dagmc_stats/DagmcFile.py
kkiesling: try this: `meshset = self._my_moab_core.get_entities_by_type_and_tag(self.root_set, types.MBENTITYSET [self.dagmc_tags['geom_dim'],self.dagmc_tags['global_id']], [[dim], ids])`

I think `dim` has to be in brackets

yqin43: @kkiesling I tried this but it still reports error `AssertionError: Only single-tag queries are currently supported.`

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,50 @@ def __set_entityset_ranges(self):
...
+        ids : (Integer) Global ID(s) of the meshset
+
+        outputs
+        -------
+        meshset : meshset of the geometry with given dimension and ids. First,
+                  dim will be checked. If dim is invalid, the root set will be
+                  returned. Then, if id is invalid (empty or not in the given
+                  dim range), all entities with the given dim will be returned.
+        """
+        if isinstance(dim, int) and dim in [0, 1, 2, 3]:
```
gonuke, line 128: Connect this to the definition of this range of valid values

```suggestion
        if isinstance(dim, int) and dim in self.entityset_types.keys():
```

dagmc_stats/DagmcFile.py
gonuke: I think this error message is probably well-formed to tell us that it won't work.  Specifically, it looks like a message @pshriwise wrote to say that this functionality is defined in C++ but not in python... sorry for the misdirection

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,50 @@ def __set_entityset_ranges(self):
...
+            dim = self.entityset_types[dim]
+        elif isinstance(dim, basestring) and dim.lower() in ['node', 'curve', 'surface', 'volume', 'nodes', 'curves', 'surfaces', 'volumes']:
```
gonuke, line 130: If allowing a string version of the dimension is going to be routine, I recommend a method that will do all the work of checking that it's valid and converting it to an integer.

gonuke, line 130: If allowing a string version of the dimension is going to be routine/common in the API, I recommend a method that will do all the work of checking that it's valid and converting it to an integer.

Note: you don't yet convert to integer, so this mode may fail

dagmc_stats/DagmcFile.py
gonuke: I think a better way to do this is to make an entityset for each dimension that contains all of the entitysets of that dimension, and then pass that as the first entry of `get_by_type_and_tag()`

gonuke (changes requested):
A few comments to clean up this new method

dagmc_stats/DagmcFile.py
kkiesling, line 130: nitpick: this line is too long for PEP8

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,52 @@ def __set_entityset_ranges(self):
...
+            return self.entityset_ranges[dim]
+
+        dim_ms = self._my_moab_core.create_meshset()
+        self._my_moab_core.add_entity(dim_ms, self.entityset_ranges[dim])
+        meshset = []
+        for id in ids:
+            meshset.extend(list(self._my_moab_core.get_entities_by_type_and_tag(dim_ms,
+                                            types.MBENTITYSET, self.dagmc_tags['global_id'], [id])))
+        # if id is not in the given dim range
+        if not meshset:
```
gonuke, line 151: I think it's best to return an empty list if there are no matches.

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,52 @@ def __set_entityset_ranges(self):
...
+            # invalid dim
+            warnings.warn('Invalid dim!')
+            return self.root_set
+
+        # if no id is passed in
+        if len(ids) == 0:
+            return self.entityset_ranges[dim]
+
+        dim_ms = self._my_moab_core.create_meshset()
+        self._my_moab_core.add_entity(dim_ms, self.entityset_ranges[dim])
```
gonuke, line 145: Do you think it makes sense to make these sets during `__init__()`?
There could then be a meshset for each dimension in a dictionary.

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,52 @@ def __set_entityset_ranges(self):
...
+                  returned. Then, if id is invalid (empty or not in the given
+                  dim range), all entities with the given dim will be returned.
+        """
+        if isinstance(dim, int) and dim in self.entityset_types.keys():
+            dim = self.entityset_types[dim]
+        #print(list(my_dict.keys())[list(my_dict.values()).index(112)])
+        elif type(dim) == str and dim.lower() in ['node', 'curve', 'surface', 'volume', 'nodes', 'curves', 'surfaces', 'volumes']:
+            dim = dim.lower()
+            if dim[-1] != 's':
+                dim = dim + 's'
```
gonuke, line 134: This `dim` needs to be converted to an integer before line 142

gonuke (changes requested):
Thanks @yqin43  - A few more ways to simplify this method

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,52 @@ def __set_entityset_ranges(self):
...
+        -------
+        meshset : meshset of the geometry with given dimension and ids. First,
+                  dim will be checked. If dim is invalid, the root set will be
+                  returned. Then, if id is invalid (empty or not in the given
+                  dim range), all entities with the given dim will be returned.
+        """
+        if isinstance(dim, int) and dim in self.entityset_types.keys():
+            dim = self.entityset_types[dim]
+        #print(list(my_dict.keys())[list(my_dict.values()).index(112)])
+        elif type(dim) == str and dim.lower() in ['node', 'curve', 'surface', 'volume', 'nodes', 'curves', 'surfaces', 'volumes']:
```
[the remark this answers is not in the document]
kkiesling: blocking/non-blocking? Just wondering if this is too clever, though I do like it.

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +109,52 @@ def __set_entityset_ranges(self):
...
+            return self.root_set
+
+        # if no id is passed in
+        if len(ids) == 0:
+            return self.entityset_ranges[dim]
+
+        dim_ms = self._my_moab_core.create_meshset()
+        self._my_moab_core.add_entity(dim_ms, self.entityset_ranges[dim])
+        meshset = []
+        for id in ids:
```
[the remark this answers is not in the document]
kkiesling: Not in the python API

dagmc_stats/DagmcFile.py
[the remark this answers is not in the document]
gonuke: I'd say blocking.... while this particular approach may appear too clever, it is important to define the valid keywords in a single place and reuse

dagmc_stats/DagmcFile.py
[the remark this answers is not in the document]
gonuke: The python API only support one tag at a time, but appear to support multiple values for that tag:

```

values : iterable of tag values
            an iterable data structure of the tag values to be matched for
            entities returned by the call (see general description for details
            on the composition of this data structure)
```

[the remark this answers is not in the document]
kkiesling: I feel like we tried this though and it didn't work.. maybe @yqin43 confirm again? Does removing the for-loop and running this one time actually work?
`self._my_moab_core.get_entities_by_type_and_tag(dim_ms, types.MBENTITYSET, self.dagmc_tags['global_id'], ids)`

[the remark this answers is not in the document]
yqin43: @gonuke @kkiesling I tried this but there is error with message '???'
```
dagmc_stats/DagmcFile.py:153: in get_meshset_by_id
    meshset = list(self._my_moab_core.get_entities_by_type_and_tag(dim_ms, types.MBENTITYSET, self.dagmc_tags['global_id'], ids))
pymoab/core.pyx:1613: in pymoab.core.Core.get_entities_by_type_and_tag
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   AssertionError

pymoab/core.pyx:1748: AssertionError
```
It seems to be working for ids with one value but not for list of ids.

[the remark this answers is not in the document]
gonuke: Hmmm?  How about 
`self._my_moab_core.get_entities_by_type_and_tag(dim_ms, types.MBENTITYSET, self.dagmc_tags['global_id'], [ids])`?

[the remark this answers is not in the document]
yqin43: @gonuke It will return this error:
```
dagmc_stats/DagmcFile.py:153: in get_meshset_by_id
    meshset = self._my_moab_core.get_entities_by_type_and_tag(dim_ms, types.MBENTITYSET, self.dagmc_tags['global_id'], [ids])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: 'pymoab.tag.Tag' object has no attribute '__getitem__'

pymoab/core.pyx:1618: TypeError
```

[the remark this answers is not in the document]
gonuke: Ok. I guess we'll leave it for now.  I looked through the pymoab tests and I don't see a test for the case I'm describing, although the code makes it look like it should work

dagmc_stats/DagmcFile.py
kkiesling: line 142 uses the string version of `dim`, as does the rest of the function. The integer version is converted to string above this.

gonuke: oh! must have misread this

tests/test_dagmc_file.py
```diff
@@ -77,6 +80,88 @@ def test_set_entityset_ranges():
...
     assert(all(test_pass))
 
+def test_set_dimension_meshset():
+    """
+    Tests different aspects of the set_dimension_meshset function
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    range = single_cube._my_moab_core.get_entities_by_type(
+                            single_cube.root_set, types.MBENTITYSET)
+    assert(set(single_cube.dim_dict.values())==set(range[-4:]))
```
yqin43, line 90: I found that the dictionary elements are not in the insertion order for adding key-value pairs to the empty dictionary, so I used set() == set() here. Or should I sort the dictionary and check if the elements are in the insertion order?

dagmc_stats/DagmcFile.py
```diff
@@ -106,6 +111,70 @@ def __set_entityset_ranges(self):
...
+        if isinstance(dim, int) and dim in self.entityset_types.keys():
+            dim = self.entityset_types[dim]
+        elif type(dim) == str and dim.lower() in all_names:
+            dim = dim.lower()
+            if dim[-1] != 's':
+                dim = dim + 's'
+        else:
+            # invalid dim
+            warnings.warn('Invalid dim!')
+            return self.root_set
```
gonuke, line 161: I think we should return nothing here.

gonuke (changes requested):
This is getting close @yqin43 - thanks for these additions.

tests/test_dagmc_file.py
```diff
@@ -77,6 +80,88 @@ def test_set_entityset_ranges():
...
     assert(all(test_pass))
 
+def test_set_dimension_meshset():
+    """
+    Tests different aspects of the set_dimension_meshset function
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    range = single_cube._my_moab_core.get_entities_by_type(
+                            single_cube.root_set, types.MBENTITYSET)
+    assert(sorted(single_cube.dim_dict.values())==list(range[-4:]))
```
kkiesling, line 90: dimlist = ['node', 'curves', 'surfaces', 'volumes']
for dim, meshset in dim_dict.values():
    if dim in dimlist:
        True
    if meshset in rangs[-4:]:
        True

kkiesling: `len(set(dim_dict.values())) == 4`
`len(set(dim_dict.keys())) == 4`

kkiesling: `sorted(dim_dict.keys()) == sorted(dim_list)`
`sorted(dim_dict.values()) == sorted(range[-4:])`

kkiesling: lots of comments- the last comment I think is the most succinct way to test the keys and values

kkiesling (commented):
I think this looks good to me. I could take it or leave it to add the new meshsets to the native_range dict.

tests/test_dagmc_file.py
```diff
@@ -77,6 +80,88 @@ def test_set_entityset_ranges():
             type_range == single_cube.entityset_ranges[set_type])
     assert(all(test_pass))
 
+def test_set_dimension_meshset():
+    """
+    Tests different aspects of the set_dimension_meshset function
+    """
+    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    range = single_cube._my_moab_core.get_entities_by_type(
```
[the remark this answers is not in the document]
gonuke: Even if it is not a syntax issue, probably good to change it for readability

tests/test_dagmc_file.py
```diff
@@ -49,7 +49,10 @@ def test_set_native_ranges():
     for i, native_range_type in enumerate(single_cube.entity_types):
         range = single_cube._my_moab_core.get_entities_by_type(
             single_cube.root_set, native_range_type)
-        test_pass[i] = (range == single_cube.native_ranges[native_range_type])
+        if native_range_type == types.MBENTITYSET:
+            test_pass[i] = (range[:-4] == single_cube.native_ranges[native_range_type])
```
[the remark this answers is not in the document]
gonuke: From a stats point of view, these sets are not part of the geometry, though, so we might not want to include them for that reason.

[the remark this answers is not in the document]
gonuke: On a related note - we may want to use a variable for this number `4` and have a comment near that variable describing why we have to skip the last 4 elements of ranges like this

gonuke (changes requested):
Thanks @yqin43 - this PR has helped us get some clarity on how we want to structure things and is getting pretty close.

tests/test_dagmc_file.py
```diff
@@ -47,9 +47,14 @@ def test_set_native_ranges():
     single_cube = ds.DagmcStats(test_env[1]['input_file'])
     test_pass = np.full(len(single_cube.entity_types), False)
     for i, native_range_type in enumerate(single_cube.entity_types):
-        range = single_cube._my_moab_core.get_entities_by_type(
+        entity_range = single_cube._my_moab_core.get_entities_by_type(
             single_cube.root_set, native_range_type)
-        test_pass[i] = (range == single_cube.native_ranges[native_range_type])
+        if native_range_type == types.MBENTITYSET:
+            # dimension meshsets are of type MBENTITYSE but not in native_ranges
+            num_dim_ms = 4
```
gonuke, line 54: I think this should be defined by the size of `self.entityset_ranges`, e.g.
```suggestion
            num_dim_ms = single_cube.entityset_ranges.size()
```

gonuke (changes requested):
One suggestion to ensure long term consistency

gonuke:
Do we want the calculation of tris per vertex in this PR? or is it for the next PR?  I thought this PR was just about the `get_mesh_id` feature.

kkiesling:
> Do we want the calculation of tris per vertex in this PR? or is it for the next PR? I thought this PR was just about the `get_mesh_id` feature.

That is supposed to be in the next PR

kkiesling (approved):
This looks good to me now!

gonuke (approved):

gonuke:
Thanks @yqin43 !

PR #84, opened 2021-07-13
Tris per vert

yqin43:
## Description
Should come after pull request #83 
Added calc_tris_per_vert() function and its corresponding test. Data frame initialization is moved to Dagmc_Query.py.

### DIFF
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -1,6 +1,7 @@
 from pymoab.rng import Range
 from pymoab import core, types
 import pandas as pd
+import numpy as np
 import warnings
 
 class DagmcQuery:
@@ -54,16 +55,16 @@ class DagmcQuery:
             tris_lst.extend(tris)
         return tris
 
-    def calc_tris_per_vert(self):
+    def calc_tris_per_vert(self, ignore_zero=True):
         """
         popoulate triangle per vertex
         """
         t_p_v_data = []
         tri_dimension = 2
-        for vertex in native_ranges[types.MBVERTEX]:
-            tpv_val = my_core.get_adjacencies(vertex, tri_dimension).size()
+        for vertex in self.dagmc_file.native_ranges[types.MBVERTEX]:
+            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(vertex, tri_dimension).size()
             if ignore_zero and tpv_val == 0:
                 continue
             t_p_v_data.append(tpv_val)
-        df_vert['t_p_v'] = np.array(t_p_v_data)
+        self._vert_data['t_p_v'] = np.array(t_p_v_data)
         return
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -75,4 +75,3 @@ class DagmcQuery:
                 continue
             t_p_v_data.append(tpv_val)
         self._vert_data['t_p_v'] = np.array(t_p_v_data)
-        return
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -4,8 +4,20 @@ import pandas as pd
 import numpy as np
 import warnings
 
+
 class DagmcQuery:
     def __init__(self, dagmc_file, meshset=None):
+        """Constructor
+
+        inputs
+        ------
+        dagmc_file : DagmcFile instance
+        meshset: the meshset on which query will be performed. The rootset will be used by default.
+
+        outputs
+        -------
+        none
+        """
         self.dagmc_file = dagmc_file
         if meshset is None:
             self.meshset = self.dagmc_file.root_set
@@ -16,7 +28,7 @@ class DagmcQuery:
         self._tri_data = pd.DataFrame()
         self._surf_data = pd.DataFrame()
         self._vol_data = pd.DataFrame()
-    
+
     def get_tris(self):
         """Get triangles of a volume if geom_dim is 3
         Get triangles of a surface if geom_dim is 2
@@ -37,27 +49,31 @@ class DagmcQuery:
         if self.meshset is None or self.meshset == self.dagmc_file.root_set:
             meshset_lst.append(self.dagmc_file.root_set)
         else:
-            dim = self.dagmc_file._my_moab_core.tag_get_data(self.dagmc_file.dagmc_tags['geom_dim'], self.meshset)[0][0]
+            dim = self.dagmc_file._my_moab_core.tag_get_data(
+                self.dagmc_file.dagmc_tags['geom_dim'], self.meshset)[0][0]
             # get triangles of a volume
             if dim == 3:
-                surfs = self.dagmc_file._my_moab_core.get_child_meshsets(self.meshset)
+                surfs = self.dagmc_file._my_moab_core.get_child_meshsets(
+                    self.meshset)
                 meshset_lst.extend(surfs)
             # get triangles of a surface
             elif dim == 2:
                 meshset_lst.append(self.meshset)
             else:
-                warnings.warn('Meshset is not a volume nor a surface! Rootset will be used by default.')
+                warnings.warn(
+                    'Meshset is not a volume nor a surface! Rootset will be used by default.')
                 meshset_lst.append(self.dagmc_file.root_set)
 
         for item in meshset_lst:
-            tris = self.dagmc_file._my_moab_core.get_entities_by_type(item, types.MBTRI)
+            tris = self.dagmc_file._my_moab_core.get_entities_by_type(
+                item, types.MBTRI)
             tris_lst.extend(tris)
         return tris_lst
 
     def calc_tris_per_vert(self, ignore_zero=True):
         """
         calculate triangle per vertex data
-        
+
         inputs
         ------
         ignore_zero : (boolean) whether or not to ignore zero tris_per_vert values
@@ -73,20 +89,25 @@ class DagmcQuery:
         if self.meshset is None or self.meshset == self.dagmc_file.root_set:
             meshset_lst.append(self.dagmc_file.root_set)
         else:
-            dim = self.dagmc_file._my_moab_core.tag_get_data(self.dagmc_file.dagmc_tags['geom_dim'], self.meshset)[0][0]
+            dim = self.dagmc_file._my_moab_core.tag_get_data(
+                self.dagmc_file.dagmc_tags['geom_dim'], self.meshset)[0][0]
             if dim == 3:
-                surfs = self.dagmc_file._my_moab_core.get_child_meshsets(self.meshset)
+                surfs = self.dagmc_file._my_moab_core.get_child_meshsets(
+                    self.meshset)
                 meshset_lst.extend(surfs)
             elif dim == 2:
                 meshset_lst.append(self.meshset)
             else:
-                warnings.warn('Meshset is not a volume nor a surface! Rootset will be used by default.')
+                warnings.warn(
+                    'Meshset is not a volume nor a surface! Rootset will be used by default.')
                 meshset_lst.append(self.dagmc_file.root_set)
         for item in meshset_lst:
-            verts.extend(self.dagmc_file._my_moab_core.get_entities_by_type(item, types.MBVERTEX))
+            verts.extend(self.dagmc_file._my_moab_core.get_entities_by_type(
+                item, types.MBVERTEX))
         verts = list(set(verts))
         for vert in verts:
-            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(vert, tri_dimension).size()
+            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(
+                vert, tri_dimension).size()
             if ignore_zero and tpv_val == 0:
                 continue
             row_data = {'vert_eh': vert, 'tri_per_vert': tpv_val}
@@ -94,4 +115,5 @@ class DagmcQuery:
         if self._vert_data.empty:
             self._vert_data = self._vert_data.append(t_p_v_data)
         else:
-            self._vert_data.set_index('vert_eh').join(self._vert_data.append(t_p_v_data).set_index('vert_eh'))
+            self._vert_data.set_index('vert_eh').join(
+                self._vert_data.append(t_p_v_data).set_index('vert_eh'))
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -7,7 +7,10 @@ import warnings
 class DagmcQuery:
     def __init__(self, dagmc_file, meshset=None):
         self.dagmc_file = dagmc_file
-        self.meshset = meshset
+        if meshset is None:
+            self.meshset = self.dagmc_file.root_set
+        else:
+            self.meshset = meshset
         # initialize data frames
         self._vert_data = pd.DataFrame()
         self._tri_data = pd.DataFrame()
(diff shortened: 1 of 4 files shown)


dagmc_stats/DagmcQuery.py
```diff
@@ -43,3 +54,17 @@ def get_tris(self):
             tris = self.dagmc_file._my_moab_core.get_entities_by_type(item, types.MBTRI)
             tris_lst.extend(tris)
         return tris
+
+    def calc_tris_per_vert(self, ignore_zero=True):
+        """
+        popoulate triangle per vertex
+        """
```
kkiesling, line 61: add input description for `ignore_zero`

dagmc_stats/DagmcQuery.py
```diff
@@ -43,3 +54,17 @@ def get_tris(self):
...
+        t_p_v_data = []
+        tri_dimension = 2
+        for vertex in self.dagmc_file.native_ranges[types.MBVERTEX]:
+            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(vertex, tri_dimension).size()
+            if ignore_zero and tpv_val == 0:
+                continue
+            t_p_v_data.append(tpv_val)
+        self._vert_data['t_p_v'] = np.array(t_p_v_data)
```
kkiesling, line 69: let's actually call this `"Triangles per vertex"`

tests/test_dagmc_query.py
```diff
@@ -68,3 +84,12 @@ def test_get_tris_dimension_incorrect():
...
     assert(all(test_pass))
+
+def test_calc_tris_per_vert():
+    """Tests part of the get_triangles_per_vertex function"""
+    three_vols = ds.DagmcStats(test_env[0]['input_file'])
+    three_vols_query = dq.DagmcQuery(three_vols)
+
+    three_vols_query.calc_tris_per_vert()
+    vertices = three_vols._my_moab_core.get_entities_by_type(three_vols.root_set, types.MBVERTEX).size()
+    assert(len(three_vols_query._vert_data['t_p_v']) == vertices)
```
kkiesling, line 95:
```suggestion
    assert(sorted(three_vols_query._vert_data['t_p_v']) == [4, 4, 4, 4, 5, 5, 5, 5])
```
That is for the single-cube.h5m geometry. This will check that the values are correct and that we have the right number of entries. But it doesn't check which verts they go to. And I think that is okay because we don't have a good way to know the vertex entity handles and how they match the number of triangles.

tests/test_dagmc_query.py
```diff
@@ -68,3 +84,12 @@ def test_get_tris_dimension_incorrect():
             if 'Meshset is not a volume nor a surface! Rootset will be used by default.' in str(w[-1].message):
                 test_pass[1] = True
     assert(all(test_pass))
+
+def test_calc_tris_per_vert():
+    """Tests part of the get_triangles_per_vertex function"""
+    three_vols = ds.DagmcStats(test_env[0]['input_file'])
```
kkiesling, line 90: this test could be made simpler by using the `single_cube` geometry.

kkiesling (changes requested):
some simple updates to the test

dagmc_stats/DagmcQuery.py
```diff
@@ -43,3 +54,17 @@ def get_tris(self):
...
+        return
```
kkiesling, line 70: unnecessary return statement here

dagmc_stats/DagmcQuery.py
kkiesling: also is this creating a new column different from the `tris_per_vert` column created in the init?

kkiesling: and do we know if these are getting put into the data frame at the correct vertex entity handle?

kkiesling: maybe a list of dictionaries: `t_p_v_data.append({'vert_eh': vertex, `tris_per_vert`: tpv_val})`

tests/test_dagmc_file.py
```diff
@@ -1,6 +1,6 @@
 from pymoab import core, types
 from pymoab.rng import Range
-import dagmc_stats.DagmcFile as ds
+import dagmc_stats.DagmcFile as df
 import dagmc_stats.DagmcQuery as dq
 import pandas as pd
```
gonuke, line 5: Don't need DagmcQuery in these tests, right?

dagmc_stats/DagmcQuery.py
```diff
@@ -43,3 +54,24 @@ def get_tris(self):
...
+        none
+        """
+        t_p_v_data = []
+        tri_dimension = 2
+        for vertex in self.dagmc_file.native_ranges[types.MBVERTEX]:
+            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(vertex, tri_dimension).size()
+            if ignore_zero and tpv_val == 0:
+                continue
+            row_data = {'vert_eh': vertex, 'tris_per_vert': tpv_val}
+            t_p_v_data.append(row_data)
```
gonuke, line 77: You need to store this data in your data frame

gonuke (changes requested):
This is coming together nicely @yqin43

kkiesling (approved):
This looks ready to me! Thanks!

tests/test_dagmc_file.py
```diff
@@ -44,7 +25,7 @@ def test_load_file_populate():
 def test_set_native_ranges():
     """Tests set_native_ranges
     """
-    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    single_cube = df.DagmcFile(test_env[1]['input_file'])
```
gonuke, line 28: discussion: I'm sure it was a prior review that resulted in this `test_env` data structure, but it seems a little odd now because you use a variable name that is strongly tied to the input file in question.

This might be a good topic for a software review meeting.... 

Some thoughts:
* should we just define the input file name in the test?
* should we use a more generic variable name, like `geometry_model` instead of `single_cube`?  Many tests might then actually work for multiple input files (which may or may not be valuable...)
* this structure creates a kind of magic number about which entry in `test_env` to use - should that be a dictionary?

tests/test_dagmc_query.py
```diff
@@ -68,3 +80,12 @@ def test_get_tris_dimension_incorrect():
             if 'Meshset is not a volume nor a surface! Rootset will be used by default.' in str(w[-1].message):
                 test_pass[1] = True
     assert(all(test_pass))
+
+def test_calc_tris_per_vert():
```
gonuke, line 84: Can we add a test for the case where a meshset is given when the `DagmcQuery` is created and confirm the expected behavior?

gonuke (changes requested):
I think there are some good design questions/discussion points here.

dagmc_stats/DagmcQuery.py
```diff
@@ -1,11 +1,18 @@
...
+import pandas as pd
+import numpy as np
 import warnings
 
 class DagmcQuery:
     def __init__(self, dagmc_file, meshset=None):
         self.dagmc_file = dagmc_file
         self.meshset = meshset
+        # initialize data frames
+        self._vert_data = pd.DataFrame()
```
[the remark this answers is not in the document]
kkiesling: We played around with this last week- I think it caused an error when we tried to add data later. I don't recall exactly the details though.

dagmc_stats/DagmcQuery.py
```diff
@@ -1,11 +1,18 @@
 from pymoab.rng import Range
 from pymoab import core, types
+import pandas as pd
+import numpy as np
 import warnings
 
 class DagmcQuery:
     def __init__(self, dagmc_file, meshset=None):
         self.dagmc_file = dagmc_file
         self.meshset = meshset
```
kkiesling, line 10: We should we probably check here `if meshset is None` and set `self.meshset` to the rootset if it is. This way we can use `self.meshset` everywhere in this class.

tests/test_dagmc_file.py
kkiesling: Something like this makes sense:
test_env = {'three_vols': 'tests/3vols.h5m', 'single_cube': 'tests/single-cube.h5m', 'pyramid': 'tests/pyramid.h5m'}

dagmc_stats/DagmcQuery.py
```diff
@@ -1,11 +1,21 @@
...
+import pandas as pd
+import numpy as np
 import warnings
 
 class DagmcQuery:
     def __init__(self, dagmc_file, meshset=None):
         self.dagmc_file = dagmc_file
-        self.meshset = meshset
+        if meshset is None:
+            self.meshset = self.dagmc_file.root_set
```
gonuke, line 11: We should be sure to document in the docstring (we need docstring) that the de facto default is the root set.

dagmc_stats/DagmcQuery.py
```diff
@@ -43,3 +53,29 @@ def get_tris(self):
...
+        for vertex in self.dagmc_file._my_moab_core.get_entities_by_type(
+                self.meshset, types.MBVERTEX):
+            tpv_val = self.dagmc_file._my_moab_core.get_adjacencies(vertex, tri_dimension).size()
+            if ignore_zero and tpv_val == 0:
+                continue
+            row_data = {'vert_eh': vertex, 'tri_per_vert': tpv_val}
+            t_p_v_data.append(row_data)
+        if self._vert_data.empty:
+            self._vert_data = self._vert_data.append(t_p_v_data)
+        else:
```
gonuke, line 80: If each query is associated with a single meshset, is it possible for this to be non-empty?

tests/test_dagmc_file.py
```diff
@@ -14,27 +13,9 @@ def test_load_file():
     """Tests loading file and check the values of all the variables that are
     set in the constructor
     """
-
-    single_cube = ds.DagmcStats(test_env[1]['input_file'])
+    single_cube = df.DagmcFile(test_env[1]['input_file'])
```
gonuke, line 16: We probably want the same changes that were maze in the query tests for this data structure.

gonuke (changes requested):

dagmc_stats/DagmcQuery.py
kkiesling: It's checking if the whole dataframe is empty. It is possible to call multiple queries that are indexed by vertex on a single meshset such as this tris per vert or roughness that would all go into this same dataframe. If one is called before the other, the dataframe may not be empty.

dagmc_stats/DagmcQuery.py
```diff
@@ -27,19 +49,71 @@ def get_tris(self):
...
+            if dim == 3:
+                surfs = self.dagmc_file._my_moab_core.get_child_meshsets(
+                    self.meshset)
+                meshset_lst.extend(surfs)
+            elif dim == 2:
+                meshset_lst.append(self.meshset)
+            else:
+                warnings.warn(
+                    'Meshset is not a volume nor a surface! Rootset will be used by default.')
+                meshset_lst.append(self.dagmc_file.root_set)
```
gonuke, line 103: It seems like we need a method (internal?) to convert the `meshset` provided by the user into a valid `meshset_lst` - then we can reuse the code here and in `get_tris`

Alternatively, we could have an internal `get_entities` that takes an entity type and a meshset that gets converted to a valid meshset_lst

gonuke (changes requested):
Important opportunity for code simplification/reuse

kkiesling (approved):
This looks good to me now. If @gonuke is happy with the reworking of the `__get_entities()` function, I think it can be merged.

tests/test_dagmc_query.py
```diff
@@ -1,68 +1,186 @@
...
+    exp = [three_vols.root_set]
+    assert(three_vols_query.meshset_lst == exp)
+
+def test_get_entities_vol():
+    """Tests the get_entities function for volume meshset
     """
-    Tests the get_tris function for volume meshset
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    vol = three_vols._my_moab_core.get_entities_by_type_and_tag(
+        three_vols.root_set, types.MBENTITYSET, three_vols.dagmc_tags['geom_dim'], [3])[0]
```
gonuke, line 43: readability/maintainability: let's use the entityset ranges we've already extracted.  This change can be used in a few other tests below- please review
```suggestion
    vol = three_vols.entityset_ranges[3][0]
```

tests/test_dagmc_query.py
```diff
@@ -1,68 +1,186 @@
...
+    exp = [surf]
+    assert(three_vols_query.meshset_lst == exp)
+
+def test_get_entities_incorrect_dim():
+    """Tests the get_entities function given incorrect dimension
+    """
+    test_pass = np.full(3, False)
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    # check if get_tris function generates warning for meshset with invalid dimension
+    verts = three_vols._my_moab_core.get_entities_by_type_and_tag(
```
gonuke, line 64: Consistency/maintainability: in the above tests, you save only the first entity from the range and use it in your query.  Here (and in the next few tests) you save the whole list and then use the first entity in your query.  It is easier for reviewers/future authors to follow if it is consistent (unless it needs to be different, which it doesn't here)

tests/test_dagmc_query.py
```diff
@@ -1,68 +1,186 @@
...
+        three_vols.root_set, types.MBENTITYSET, three_vols.dagmc_tags['geom_dim'], [0])
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter('always')
+        three_vols_query = dq.DagmcQuery(three_vols, meshset=verts[0])
+        obs_tris = three_vols_query.get_tris()
```
gonuke, line 69: Doesn't the error come before this line?  This is a `get_entities` test, so should not need to call `get_tris`.  That test is below.

tests/test_dagmc_query.py
```diff
@@ -1,68 +1,186 @@
...
         three_vols.root_set, types.MBENTITYSET, three_vols.dagmc_tags['geom_dim'], [0])
     with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter('always')
         three_vols_query = dq.DagmcQuery(three_vols, meshset=verts[0])
         obs_tris = three_vols_query.get_tris()
+        if len(w) == 1:
+            test_pass[0] = True
+            if 'Meshset is not a volume nor a surface! Rootset will be used by default.' in str(w[-1].message):
+                test_pass[1] = True
+    assert(all(test_pass))
```
gonuke, line 136: Should you also test that you get all the tris because its now called on the rootset?

tests/test_dagmc_query.py
```diff
@@ -1,68 +1,186 @@
...
+    """
+    test_pass = np.full(2, False)
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    # check if get_tris function generates warning for meshset with invalid dimension
+    verts = three_vols._my_moab_core.get_entities_by_type_and_tag(
+        three_vols.root_set, types.MBENTITYSET, three_vols.dagmc_tags['geom_dim'], [0])
+    with warnings.catch_warnings(record=True) as w:
         warnings.simplefilter('always')
+        three_vols_query = dq.DagmcQuery(three_vols, meshset=verts[0])
+        three_vols_query.calc_tris_per_vert()
```
gonuke, line 183: Also test that this gives an answer for every vert in the rootset?

gonuke (changes requested):
A few more changes, almost all in support of long term comprehension of the code.  It is all algorithmically correct, but a few places where inconsistencies make it harder to follow.

dagmc_stats/DagmcQuery.py
```diff
@@ -1,12 +1,54 @@
...
+        outputs
+        -------
+        none
+        """
         self.dagmc_file = dagmc_file
-        self.meshset = meshset
-    
+        self.meshset_lst = []
+        if meshset is None:
+            self.__get_entities(self.dagmc_file.root_set)
```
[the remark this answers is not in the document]
kkiesling: probably not, but you could argue we should call `__get_entities()` for consistency. I think it can go either way as far as maintainability goes.

kkiesling (commented):
I think this is ready....

gonuke:
Thanks for all the work to make some well-written a reusable capability here @yqin43

PR #85, opened 2021-09-24
Tri data

yqin43:
## Description
Added the calc_area_triangle function, the calc_triangle_aspect_ratio function and their corresponding tests.


dagmc_stats/DagmcQuery.py
```diff
@@ -99,5 +138,62 @@ def calc_tris_per_vert(self, ignore_zero=True):
...
+        none
+
+        outputs
+        -------
+        none
+        """
+        t_a_r_data = []
+        tris = self.get_tris()
+        for tri in tris:
+            side_lengths = list(self.get_tri_side_length(tri).values())
```
kkiesling, line 159: is it redundant to do `list( )` on this? I think `.values()` on a dictionary already returns a list.

tests/test_dagmc_query.py
```diff
@@ -62,7 +62,7 @@ def test_get_entities_incorrect_dim():
     vert = three_vols.entityset_ranges['nodes'][0]
     with warnings.catch_warnings(record=True) as w:
         warnings.simplefilter('always')
-        three_vols_query = dq.DagmcQuery(three_vols, meshset=vert)
+        three_vols_query = dq.DagmcQuery(three_vols, vert)
```
kkiesling, line 65: curious: what is the reason for removing `meshset=` here?

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
...
     assert(all(test_pass))
+
+def test_calc_area_triangle_vol():
+    """Tests part of the calc_area_triangle function
+    """
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    vol = three_vols.entityset_ranges['volumes'][0]
+    three_vols_query = dq.DagmcQuery(three_vols, vol)
+    three_vols_query.calc_area_triangle()
+    assert(all(three_vols_query._tri_data['area']) == all(np.full(12, 50)))
```
kkiesling, line 192: 12 and 50 are magic numbers in this line. Can you add variables to explain what they are? I think 12 is the number of triangles and 50 is the expected aspect ratio?

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
             three_vols.root_set, types.MBVERTEX))
         test_pass[2] = (len(three_vols_query._vert_data['tri_per_vert']) == exp_tpv_len)
     assert(all(test_pass))
+
```
kkiesling, line 184: PEP8: need an additional blank line here

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
...
+    vol = three_vols.entityset_ranges['volumes'][0]
+    three_vols_query = dq.DagmcQuery(three_vols, vol)
+    three_vols_query.calc_area_triangle()
+    assert(all(three_vols_query._tri_data['area']) == all(np.full(12, 50)))
+
+
+def test_calc_area_triangle_surf():
+    """
+    Tests part of the calc_area_triangle function
+    """
```
kkiesling, line 198:
```suggestion
    """Tests part of the calc_area_triangle function
    """
```

just a nitpick for consistency

tests/test_dagmc_query.py
kkiesling: (This comment applies to all the tests where such numbers appear)

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
...
+
+
+def test_calc_triangle_aspect_ratio_vol():
+    """Tests part of the calc_triangle_aspect_ratio function
+    """
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    vol = three_vols.entityset_ranges['volumes'][0]
+    three_vols_query = dq.DagmcQuery(three_vols, vol)
+    three_vols_query.calc_triangle_aspect_ratio()
+    exp = (10*10*10*np.sqrt(2))/(8*5*np.sqrt(2)*5*np.sqrt(2)*(10-5*np.sqrt(2)))
```
kkiesling, line 245: There are a lot of magic numbers here- can you explain what these are (use variable names) and also list the formula you are using. (And do so in the rest of the tests)

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
...
+
+def test_calc_triangle_aspect_ratio_rootset():
+    """Tests the calc_triangle_aspect_ratio function given the rootset
+    """
+    three_vols = df.DagmcFile(test_env['three_vols'])
+    three_vols_query = dq.DagmcQuery(three_vols, three_vols.root_set)
+    three_vols_query.calc_triangle_aspect_ratio()
+    exp_tpv_len = len(three_vols._my_moab_core.get_entities_by_type(
+        three_vols.root_set, types.MBTRI))
+    assert(len(three_vols_query._tri_data['aspect_ratio']) == exp_tpv_len)
```
kkiesling, line 268: Can you add a note about why we are only testing the length of the data here rather than the values.

kkiesling: related to this- is there a different geometry file in this test suite that could be used so we could also test the values? Or maybe we should make one that doesn't include the sphere so that this can be tested more exactly.

kkiesling (changes requested):
Some comments on the tests. I think this is fairly straight forward.

All the tests make me wonder if we can take advantage of some pytest parameterization. I don't think that is a necessary change for this PR, but it might be interesting to discuss in s/w at some point.

dagmc_stats/DagmcQuery.py
```diff
@@ -99,5 +138,62 @@ def calc_tris_per_vert(self, ignore_zero=True):
...
+            s = .5*(sum(side_lengths))
+            top = np.prod(side_lengths)
+            bottom = 8*np.prod(s-side_lengths)
+            t_a_r = top/bottom
+            row_data = {'tri_eh': tri, 'aspect_ratio': t_a_r}
+            t_a_r_data.append(row_data)
+        if self._tri_data.empty:
+            self._tri_data = self._tri_data.append(t_a_r_data)
+        else:
+            self._tri_data.set_index('tri_eh').join(self._tri_data.append(t_a_r_data).set_index('tri_eh'))
```
gonuke, line 169: We have this same pattern in each method.  Should it be a small method of its own?
```
def update_tri_data(self, new_data)
     if self._tri_data.empty:
         self._tri_data = self._tri_data.append(new_data)
     else:
         self._tri_data.set_index('tri_eh').join(self._tri_data.append(new_data).set_index('tri_eh'))
```

tests/test_dagmc_query.py
```diff
@@ -182,3 +181,109 @@ def test_calc_tris_per_vert_dimension_incorrect():
...
+
+
+def test_calc_triangle_aspect_ratio_incorrect():
```
gonuke, line 271: We should think carefully about whether we need to test all of these `*_incorrect()` tests.  They don't really test anything different from the `get_tris_incorrect()` combined with the respective `*_rootset()` method.  I don't think they improve overall testing of the code and almost certainly don't improve code coverage.

gonuke (changes requested):
Thanks @yqin43 ... these new updates are getting cleaner and cleaner thanks to the investment in the infrastructure.

dagmc_stats/DagmcQuery.py
gonuke: On a related note...  I don't think any of your tests are designed to test this update capability.  They all start from empty `DagmcQuery` objects.  Perhaps we need some new tests that:
• call the same query multiple times on the same meshset => the database should not grow
• call the same query multiple times on different mesh sets => the database should grow appropriately 
• call different queries on the same meshsets
• call different queries on different meshsets

(Not sure what the expected behavior is for the last two)

kkiesling: regarding these tests:

* call the same query multiple times on the same meshset => the database should not grow --> we need to add a check to make sure we are not adding a second column (currently it does)
* call the same query multiple times on different mesh sets => the database should grow appropriately --> you can't call multiple queries on different meshsets. Each set of meshsets gets its own query object so this action isn't a valid test. It's equivalent to calling a single query each on two different query objects.
* call different queries on the same meshsets --> this is what you probably meant in the last point. One query object with multiple different queries should grow (one test exists now to show that)
* call different queries on different meshsets --> this is the same as point 2 above

dagmc_stats/DagmcQuery.py
```diff
@@ -99,5 +142,81 @@ def calc_tris_per_vert(self, ignore_zero=True):
         if self._vert_data.empty:
             self._vert_data = self._vert_data.append(t_p_v_data)
         else:
-            self._vert_data.set_index('vert_eh').join(
-                self._vert_data.append(t_p_v_data).set_index('vert_eh'))
+            self._vert_data.set_index('vert_eh').join(self._vert_data.append(t_p_v_data).set_index('vert_eh'))
```
kkiesling, line 145: nitpick: check the PEP8 line length on this line

kkiesling: Also, this section needs to be updated to use the new `update` function

dagmc_stats/DagmcQuery.py
```diff
@@ -99,5 +142,81 @@ def calc_tris_per_vert(self, ignore_zero=True):
...
+
+    def update_tri_data(self, new_data):
```
kkiesling: I think that this could be a private function instead. It's not useful for a user I think..

dagmc_stats/DagmcQuery.py
```diff
@@ -99,5 +142,81 @@ def calc_tris_per_vert(self, ignore_zero=True):
...
+        ------
+        none
+
+        outputs
+        -------
+        none
+        """
+        if 'aspect_ratio' in self._tri_data:
+            warnings.warn(
+                    'Triangle aspect ratio already exists. Calc_triangle_aspect_ratio() will not be called.')
```
kkiesling, line 179: check line length (check throughout the whole file)

tests/test_dagmc_query.py
```diff
@@ -152,33 +157,55 @@ def test_calc_tris_per_vert_surf():
...
+    three_vols_query.calc_triangle_aspect_ratio()
+    exp_tar = (10*10*10*np.sqrt(2))/(8*5*np.sqrt(2)*5*np.sqrt(2)*(10-5*np.sqrt(2)))
+    np.testing.assert_almost_equal(list(three_vols_query._tri_data['aspect_ratio']), list(np.full(2, exp_tar)))
+    three_vols_query.calc_area_triangle()
+    np.testing.assert_almost_equal(list(three_vols_query._tri_data['area']), list(np.full(2, 50)))
 
 
-def test_calc_tris_per_vert_dimension_incorrect():
-    """Tests the calc_tris_per_vert function given incorrect dimension
+def test_duplicate_tpv():
```
kkiesling, line 194: thought: should we be testing this warning/duplicate behavior for every single `calc_X` function? Or is that too redundant? I can imaging a case where one calc_X function accidentally has the lines removed and then it doesn't function properly but it is not caught in the tests.

kkiesling (commented):
a few additional quick comments

tests/test_dagmc_query.py
```diff
@@ -152,33 +158,60 @@ def test_calc_tris_per_vert_surf():
...
+    three_vols_query.calc_triangle_aspect_ratio()
+    exp_tar = (10*10*10*np.sqrt(2))/(8*5*np.sqrt(2)
+                                     * 5*np.sqrt(2)*(10-5*np.sqrt(2)))
+    np.testing.assert_almost_equal(
+        list(three_vols_query._tri_data['aspect_ratio']), list(np.full(2, exp_tar)))
+    three_vols_query.calc_area_triangle()
+    np.testing.assert_almost_equal(
+        list(three_vols_query._tri_data['area']), list(np.full(2, 50)))
+
+@pytest.mark.parametrize("function,message", [('calc_tris_per_vert()','Tri_per_vert already exists. tris_per_vert() will not be called.')])
```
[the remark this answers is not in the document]
kkiesling: This particular parameterize is not in response to #86 but in response to making sure we have every warning message appropriately tested for each in each of the calc functions. The rest of the parameterization can come later.

kkiesling:
awesome additions!

PR #95, opened 2022-02-09
calc_tris_per_surf and its test

yqin43:
## Description
Calc_tris_per_surf and its corresponding test.


gonuke (approved):
This looks good to me!  I notice that we don't have a way for the user to query this data once we calculate it.

PR #99, opened 2022-02-21
Surfs per vol

yqin43:
## Description
Calc_surfs_per_vol() and its corresponding test

### DIFF
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -53,7 +53,7 @@ class DagmcQuery:
         -------
             none
         """
-        # allow mixed list of surfaces and volumes that create a
+        # allow mixed list of surfaces and volumes and create a
         # single list of all the surfaces together
         if type(meshset) != list:
             # convert single item to list for iterating
@@ -186,7 +186,7 @@ class DagmcQuery:
         return side_lengths
 
     def calc_tris_per_vert(self, ignore_zero=True):
-        """calculate triangle per vertex data
+        """Calculate triangle per vertex data
 
         inputs
         ------
@@ -213,7 +213,7 @@ class DagmcQuery:
         self.__update_vert_data(t_p_v_data)
         
     def calc_tris_per_surf(self):
-        """calculate triangle per surface data
+        """Calculate triangle per surface data
 
         inputs
         ------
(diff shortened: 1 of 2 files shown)


dagmc_stats/DagmcQuery.py
```diff
@@ -77,6 +80,34 @@ def __get_entities(self, meshset):
...
+            dim = self.dagmc_file._my_moab_core.tag_get_data(
+                self.dagmc_file.dagmc_tags['geom_dim'], m)[0][0]
+            # get volumes
+            if dim == 3:
+                self.vols.append(m)
+
+        self.vols = list(set(self.vols))
+        # if no items in the meshset list is a volume,
+        # then use rootset by default instead
+        if len(self.meshset_lst) == 0:
```
gonuke, line 107:
```suggestion
        if len(self.volts) == 0:
```

dagmc_stats/DagmcQuery.py
```diff
@@ -77,6 +80,34 @@ def __get_entities(self, meshset):
...
+            warnings.warn('Specified meshset(s) are not volumes. Rootset will be used by default.')
+            self.vols = self.dagmc_file.entityset_ranges['volumes']
```
gonuke, line 109: This is now more complicated.... if the meshset only contains surfaces, then we should leave this empty.  We don't want the list of volumes to be inconsistent with the list of surfaces.

dagmc_stats/DagmcQuery.py
```diff
@@ -204,6 +235,28 @@ def calc_tris_per_surf(self):
...
+            none
+        """
+        if 'surf_per_vol' in self._surf_data:
+            warnings.warn('Surf_per_vol already exists. ' +
+                          'calc_surfs_per_vol() will not be called.')
+            return
+        s_p_v_data = []
+        for vol in self.vols:
+            num_vols = self.dagmc_file._my_moab_core.get_child_meshsets(vol).size()
+            row_data = {'vol_eh': vol, 'surf_per_vol': num_vols}
```
gonuke, line 256:
```suggestion
            num_surfs = self.dagmc_file._my_moab_core.get_child_meshsets(vol).size()
            row_data = {'vol_eh': vol, 'surf_per_vol': num_surfs}
```

gonuke (changes requested):
A couple of little changes, but maybe we should discuss what the correct behavior is when there are no volumes in the meshset.

gonuke (changes requested):
Just one small change.

gonuke (approved):
Thanks for the work on this @yqin43 as we rethought some of the model on how we use the meshset

PR #101, opened 2022-04-21
add_tag

### LINKED ISSUE #94: Update add_tags for new design

yqin43:
## Description
Added add_tag() function and its corresponding test

## Behavior
How to delete the tag after adding one during the test needs to be discussed: creating a new tag entity handle and then deleting the tag with the newly created tag eh works but the new eh is different form the tag_eh created when first adding the tag.

Fixes #94

### DIFF
--- a/dagmc_stats/DagmcQuery.py
+++ b/dagmc_stats/DagmcQuery.py
@@ -608,12 +608,30 @@ class DagmcQuery:
         -------
             none
         """
+        # the default tag size is 1, else is len(dic_vals)
+        tag_size = 1
+        
+        # check if lengths of values in the tag_dic are consistent
+        if tag_dic is not None:
+            # integer elements in tag_dic.values()
+            int_val = list(filter(lambda val: type(val) is not list, tag_dic.values()))
+            # list elements in tag_dic.values()
+            lst_val = [val for val in tag_dic.values() if val not in int_val]
+            if len(lst_val) != 0:
+                tag_size = len(lst_val[0])
+
+            if len(int_val) != 0 and len(lst_val) != 0 or \
+                    len(lst_val) != 0 and not all(len(lst_val) == tag_size for val in tag_dic.values()):
+                warnings.warn('lengths in tag_dic values are not consistent!')
+                return
+
         # create the tag handle
         tag_eh = \
-            self.dagmc_file._my_moab_core.tag_get_handle(tag_name, size=1,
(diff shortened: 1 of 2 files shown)


yqin43:
There is the error
'AssertionError: Incorrect data length
for
dagmc_stats/DagmcQuery.py:636: in add_tag
    self.dagmc_file._my_moab_core.tag_set_data(tag_eh, tag_dic.keys(), tag_dic.values())'

when running pytest3 but this error does not exist in pytest2.
I am not sure if the problem rises that adding all the entries into the tag data is not supported by python3.

dagmc_stats/DagmcQuery.py
```diff
@@ -593,3 +593,48 @@ def __calc_tri_roughness(self):
...
+        if tag_dic is not None:
+            # integer elements in tag_dic.values()
+            int_val = list(filter(lambda val: type(val) is not list, tag_dic.values()))
+            # list elements in tag_dic.values()
+            lst_val = [val for val in tag_dic.values() if val not in int_val]
+            if len(lst_val) != 0:
+                tag_size = len(lst_val[0])
+
+            if len(int_val) != 0 and len(lst_val) != 0 or \
+                    len(lst_val) != 0 and not all(len(lst_val) == tag_size for val in tag_dic.values()):
```
gonuke, line 624: This seems more complicated than we need.  If I understand correctly, `tag_dic.values()` has two valid options.  They are either all scalars, or all lists of the same size.

Can we just check those criteria, one at a time?
1. count the number of entries, E, that are lists (it should either be 0 or the same as the length of the values())
2. if 0 < E < N - error/warn
3. if E==0:  get the length of the first entry and check that the remainder are all that length

gonuke: This currently probably does the same thing, but it is not clear to read/follow.

gonuke (changes requested):
Just some changes to how the logic is laid out to make it more maintainable.

dagmc_stats/DagmcQuery.py
yqin43: I think in this way the situation where the list entries are of different lengths (e.g. ([1,2,3], [1,2], [1,2]) are not considered. I would still prefer the current way.

gonuke (approved):
Thanks @yqin43 - this looks good to me.

